PR #94, opened 2014-12-09
Add option to override default resolution for Cairo backends

darwindarak:
This adds the keyword argument `dpi` to the Cairo backends, for example:

``` julia
draw(PNG("default.png", 1inch, 1inch), p)
draw(PNG("hi_res.png", 1inch, 1inch, dpi=300), p)
```

will result in a 96x96 `default.png` and a 300x300 `hi_res.png`

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -761,17 +761,9 @@ function draw(img::Image, prim::LinePrimitive)
     end
 end
 
-
-#function get_layout_size(img::PNG)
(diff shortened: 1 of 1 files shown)


bjarthur:
while the `dpi` keyword does indeed change the resulting pixel size of the image when inches are used to specify the size, the resulting PNG file still has 72 dpi encoded as the resolution, and so Preview on OS X, for example, displays it as the wrong size on the screen.  anyone know how to fix this?

PR #108, opened 2015-02-26
Modernize examples and use them as tests

jiahao:
This PR updates the examples in the `examples/` directory so that they run on
modern versions of Julia and Compose. Additionally, tests have been added which
run all the examples and compares the generated output with cached output.

I've also included a Travis configuration file so that if you want to flip the
switch, `Compose.jl` can start to have CI tests and coverage data.

Additionally, there are several spelling corrections, formatting adjustments,
and an update to the license file.

Caveat: I've had to guess what some of the examples do in order to modernize
them.

Implementation note:

The tests do not use `Base.Test`. Instead, `runtests.jl` keeps track of any
files that differ from the cached output and outputs a list of different files.
If `diff` is available locally, also prints out the diff to facilitate
comparison. The files that differ are not automatically deleted from the
`test` subdirectory where they are generated.


shashi:
:100:

dcjones:
Awesome, thanks!

I'm going to miss all my misspellings though. I'll try to add more in future commits.

jiahao:
It's more fun to cöopt obsolete diacritics :)

PR #109, opened 2015-03-03
Use relative paths to get package location

swt30:
Using `Pkg.dir("Compose")` to get the path to the Compose installation won't work if Compose is installed outside the standard package location (for example, if it's been installed system-wide for all users). This PR changes instances of `Pkg.dir("Compose")` to use relative paths with `@__FILE__` instead to access other files in the package. See JuliaLang/julia#8679

### DIFF
--- a/doc/make.jl
+++ b/doc/make.jl
@@ -1,4 +1,4 @@
 
 using Judo
-Judo.collate("Compose", template=Pkg.dir("Compose", "doc", "template"))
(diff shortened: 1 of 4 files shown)


shashi:
Just adding a note: before tagging we need to confirm that it doesn't break precompilation

yuyichao:
I believe all of the changes in the PR are included in #127...

swt30:
Cool, glad that got done.

PR #111, opened 2015-03-14
Replace "&" in text with "&amp;" to prevent Cairo errors.

glesica:
When rendering to a file, a `text(...)` element that contained an ampersand would cause the following error:
```
(process:40123): Pango-WARNING **: pango_layout_set_markup_with_accel: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;
```
I've fixed the issue by simply replacing "&" with "\&amp;" when the element is rendered.


timholy:
Maybe a test illustrating the problem (and then ensuring it stays fixed) would help get this merged?

glesica:
@timholy added an example, which gets tested automatically. The example won't run successfully on master but will with the fix.

glesica:
@timholy ping :-)

timholy:
Thanks!

bjarthur:
@glesica @timholy there is a [discussion](https://github.com/GiovineItalia/Compose.jl/issues/235) to revert this commit.  as the original submitter and the merger, could you please voice an opinion?

glesica:
@bjarthur I don't love the idea of fixing it with documentation alone since it seems to me that you shouldn't have to care which backend you're using when you draw something (which would be the case if you had to manually escape your strings). That being said, as noted, this fix only applies to one symbol on one backend, which isn't great either. What if we applied escaping to all strings on backends that require it?

bjarthur:
is there an easy way to escape strings?

glesica:
Well if we can get away with just URL encoding the strings, then it's
easy (https://www.rosettacode.org/wiki/URL_encoding#Julia). Someone
could test to see if this works for the SVG and Cairo backends. If you
give me a day or two I would be willing to see how hard this would be to
fix. If it ends up being a pain, then I'd be 100% on board with just
reverting the commit and documenting the issue properly.

bjarthur:
`URIParser.escape("foo & bar")` returns `"foo%20%26%20bar"` when i think we want `"foo &#162; bar"`

glesica:
Yeah, I bet there's some code or a library for escaping HTML though, after all, isn't there a whole web framework for Julia? Seems like it will just require a little research (maybe). We'll see.

glesica:
So what about just grabbing this function (it's MIT licensed) and running strings through that for the SVG and Cairo backends? @bjarthur 

https://github.com/JuliaWeb/HttpCommon.jl/blob/master/src/HttpCommon.jl#L113

lobingera:
The problem goes actually a little bit further: The cairobased backend uses the pango library to render text and this supports markup of text (see: https://github.com/JuliaGraphics/Cairo.jl/blob/master/samples/sample_pango_text.jl ; https://github.com/JuliaGraphics/Cairo.jl/blob/master/samples/sample_pango_text.png ; https://developer.gnome.org/pango/unstable/PangoMarkupFormat.html) while the SVG has another way if markup (see tspan: https://www.w3.org/TR/SVG/text.html#TextElement)

So, any single/simple function to text escaping will be a 'hack' rather.

bjarthur:
@glesica the function linked to replaces all `&` with  `&amp;` so a user couldn't, for example, display the symbol for a british pound by using `&#163;`.  http://www.ascii.cl/htmlcodes.htm

@lobingera thanks for the links.  i had no idea it was soo complicated, and versatile!

i think reverting the commit and adding docs is the best route forward.  what do you think?

glesica:
Yeah, that seems reasonable.

bjarthur:
great.  i'm working on a PR...

PR #119, opened 2015-04-22
Fixes v0.4 Tuple{A} syntax & updates test results

sebastiang:
The two issues are independent so I left them as two commits.
The first just updates tuple type args to the new v0.4 format using the `@compat` macro.
The second updates test results for an apparent change from preferring `fill="none"` to `fill="#000000" fill-opacity="0.000"`. The latter is more verbose, and I suppose it's possible it's due to the `Nothing` -> `Void` rename in v0.4, but it doesn't look likely.

### DIFF
--- a/src/container.jl
+++ b/src/container.jl
@@ -419,7 +419,7 @@ function drawpart(backend::Backend, root_container::Container)
     properties = Array(Property, 0)
 
     # collect and sort container children
-    container_children = Array((Int, Int, Container), 0)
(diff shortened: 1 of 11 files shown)


dcjones:
Thanks!

dcjones:
I sincerely enjoy that whenever we refer to julia macros, it pings someone like @compat who made a github account five years ago and never used it.

sebastiang:
Brilliant. He looks pretty chill about the whole thing fortunately.

PR #127, opened 2015-07-12
Various path fixes

yuyichao:
* Use `@__FILE__` instead of `Pkg.dir`
* Close file handle after reading the data
* Ignore test pdf output
* Support loading optional dependencies from LOAD_PATH


yuyichao:
Just noticed that this is mostly https://github.com/dcjones/Compose.jl/pull/109 (plus some other fixes).

Can either of these be merged?

yuyichao:
I starts a travis run on my account and at least the package loads with no problems.

PR #128, opened 2015-07-13
Whitespace fixes for Julia 0.4

jiahao:


src/patchwork.jl
```diff
@@ -193,8 +193,8 @@ function draw(img::Patchable, prim::LinePrimitive)
 
     paths = make_paths(prim.points)
     if length(paths) > 1
-        addto(Elem(:svg, :g)
-              [Elem(:svg, :path, fill="none", d=svg_fmt_path(path, true))
+        addto(Elem(:svg, :g)[
```
shashi: well... that does not look correct either way. The path should go inside the :g tag. will fix this.

shashi:
This was fixed.

PR #130, opened 2015-07-14
Fix patchwork naming issue on OSX

shashi:
related to https://github.com/JuliaLang/julia/issues/9079

WTF!!!!!


jiahao:
This still doesn't fix the problem for me:

```
julia> using Compose #master, julia release-0.3 head

julia> Compose.Patchable
ERROR: Patchable not defined

julia> Pkg.installed("Patchwork")
v"0.1.6"

julia> Compose.isinstalled("Patchwork", v"0.1.2")
true
```

shashi:
This is not merged yet. You need to checkout the shashi-osx-fix branch.

jiahao:
> This is not merged yet. You need to checkout the shashi-osx-fix branch.

Yes, I applied this patch and i still doesn't work for me.

jiahao:
Ok never mind I forgot I had Compose precompiled on my release branch.

It does work for me now.

shashi:
I checked again, I did rename it properly with git. Can you check if
src/patchwork.jl isn't lurking? And it worked for Alan...

shashi:
Oh.

jiahao:
For the record, the problem is fixed so long as `Compose` is not precompiled. Even with this patch, the `Patchable not defined` error is thrown if `Compose` is precompiled.

shashi:
Oh wow.

PR #131, opened 2015-07-20
WIP: Replaces compose measures with Measures.jl

dcjones:
We're switching Compose to a new and improved measures system that lives in [Measures.jl](https://github.com/dcjones/Measures.jl).

I think all that remains to do is update the pgf backend, document Measures.jl a little, and put it in METADATA.

@darwindarak If you want to take a crack at this, it would be welcomed, otherwise I can handle it.

The important changes:
  * Points are now tuples. So instances of `point.x.abs` and `point.y.abs` need to be replaced with `point[1].value` and `point[2].value`, respectively. And points are constructed like `(x, y)` instead of `Point(x, y)`.
  * Bounding boxes are generalized to n dimensions. `box.x0`, `box.y0` becomes `box.x0[1]`, `box.x0[2]`, respectively. And `box.width`, `box.height` becomes `box.a[1]`, `box.a[2]` respectively.
  * There's no longer a `Measure(...)` constructor, so these should be generally be constructed by multiplication like `value*mm` or `value*cx`.


darwindarak:
Sure thing, I'll take a look at it tonight

shashi:
:clap:

rohitvarkey:
:clap:

PR #132, opened 2015-07-21
WIP^2: Update PGF backend for new Measures.jl

darwindarak:
It seems to work with the Sierpinski triangle example.  I might not have time to run comprehensive tests until this weekend though.


darwindarak:
The backend can now run all the test cases in `examples`.  They all seem to match the results of the SVG backend.  The golden rectangle output results looks wrong across all backends, but that's probably not related to the change to the new measures (cc @dcjones).

dcjones:
This looks good, thanks!

PR #137, opened 2015-08-21
Switch to Colors package (do not merge yet)

timholy:
This switches to the new Colors package, as discussed in https://github.com/JuliaLang/Color.jl/issues/101. Posting for discussion. Do not merge until C-day, tentatively scheduled for Tuesday, Aug 25.

PR #138, opened 2015-08-23
Switch to Colors package (do not merge yet)

timholy:
Supercedes #137


timholy:
Out of curiosity, is Travis testing off deliberately? I'm asking because of the results I'm getting in https://travis-ci.org/timholy/Dummy.jl/jobs/77092557

timholy:
(The tests pass locally.)

PR #142, opened 2015-08-30
Fix introspect on 0.4

timholy:


dcjones:
Thanks for this, and all the precompiles. :bouquet:

PR #143, opened 2015-09-01
Precompile some functions for faster startup

timholy:
Together with https://github.com/dcjones/Gadfly.jl/pull/673, this drops time-to-plot in approximately half on my machine. See discussion in https://github.com/JuliaLang/julia/issues/12897.


dcjones:
Thanks for doing this! The one issue with this approach is that there are few functions in this list that defined only if Fontconfig is installed, other defined only if Fontconfig is not installed, and similarly for Cairo. I commented these out, but it makes regenerating this file a little painful.

timholy:
If there's a good way to test for those functions, I can add this logic to SnoopCompile.

Issue #139, opened 2015-08-27
Precompilation causes Compose to not notice if Cairo is later installed


jwmerrill:
I installed Gadfly, used it once, and then installed Cairo so that I could draw to PNGs. Even after I installed Cairo, I was getting:

```
julia> using Gadfly

julia> PNG("myplot.png", 12cm, 6cm)
ERROR: Cairo must be installed to use the PNG backend.
 in PNG at /Users/jwm/.julia/v0.4/Compose/src/Compose.jl:137
```

I was able to fix this problem by removing all the cache files in `~/.julia/lib/v0.4/` and then running `using Gadfly` again, which recompiled everything.

dcjones:
Particularly with precompilation, there's no good way to handle optional dependencies in julia packages. It's a major deficiency, but one that will be fixed eventually. Until then, I don't think there's is much I can do about this except tell people to delete their `Gadfly.ji`.

jwmerrill:
It's safer to tell people to delete everything in the cache. I initially tried just deleting Gadfly.ji, but that doesn't clear Compose.ji, and ends up leaving things in an inconsistent state.

dcjones:
Ok, there's at least a warning now.

timholy:
I suspect that things get a lot more robust if you put `if ... && __precompile__()` at the top of each module that Gadfly uses.

dcjones:
Would that help with optional dependencies? Cairo already has a `__precompile__()`.

timholy:
More likely the "inconsistent state," which I'm guessing must be something like https://github.com/dcjones/Gadfly.jl/pull/667#issuecomment-133991074. Optional dependencies probably wouldn't be detect unless it has to recompile Gadfly/Compose itself. (By deleting Compose.ji, though, it should have triggered a recompile of both Compose and Gadfly.)

brian-j-smith:
What do you make of this ``Cairo must be installed`` error occurring on the [julialang.org package tests](http://pkg.julialang.org/?pkg=Gadfly&ver=0.4) but not on Travis?

This is happening too for my Mamba package, which uses Gadfly.  Adding an explicit ``using Cairo`` before ``using Gadfly`` to my package does not seem to help.  Testing on my local machines is fine, so I'm at a loss for how to fix the julialang.org failure.  Any help with that would be greatly appreciated.  Thanks!

timholy:
[...]

PR #146, opened 2015-09-03
Enable precompilation.

brian-j-smith:
Enable precompilation in the top-level module file in response to #139.


dcjones:
Thanks!

PR #148, opened 2015-09-08
Add tags to Context and Forms

timholy:
This allows individual objects to be "named." By walking the tree, it becomes possible to find and extract the named object again later. Useful in a variety of contexts (no pun intended), but particularly for situations where you want to know "at what coordinates did object x end up?"

### DIFF
--- a/src/container.jl
+++ b/src/container.jl
@@ -48,6 +48,8 @@ type Context <: Container
     # is preferable to another.
     penalty::Float64
 
+    tag::Symbol
+
     function Context(x0=0.0w,
                      y0=0.0h,
                      width=1.0w,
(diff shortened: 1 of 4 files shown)


timholy:
If I don't hear otherwise, I'll merge in ~24 hours.

dcjones:
This is nice. I'm wondering if this shouldn't replace the `svgid` property or be left as purely a debugging mechanism.

timholy:
Glad you like it. Obviously feel free to take things in whatever direction you want. AFAICT, the `svgid` property is "categorical," whereas this allows you to be "idiosyncratic." Obviously you could use tagging to give categorical names, but I can imagine times where you might want both. I.e., "this is a plotpanel, and this specific panel's name is :Bob."

PR #145, opened 2015-09-02
Overload showcompact for command-line display of tree structure

timholy:
Sorry to keep bombing you...more introspection.

The `tomato_bisque` example looks like
```jl
Context(Context(f,R),Context(f,C))
```
and
```jl
p = Gadfly.plot(x=rand(10^6), Gadfly.Geom.histogram(bincount=100));
```
looks like
```jl
julia> showcompact(ctx)
Context(Context(jsip,svgc,3x2 Table:
  [Context(AdhocContainerPromise)],[Context(jsc,Context(Context(Context(Context(svga,lw,Context(s,f,svgc,FORM))),svgc)),Context(Context(fo,svgc,s,Context(jsc,jsc,jsc,f,Context(svgc,f,SP),lw,so,s,R),Context(Context(jsc,jsc,jsc,jsc,jsc,svgc,f,R),Context(jsc,jsc,jsc,f,R)),Context(jsc,jsc,jsc,f,Context(svgc,f,SP),lw,so,s,R))),Context(Context(Context(jsc,jsc,svga,svgc,sd,lw,s,v,L),Context(svgc,sd,lw,s,L)),Context(Context(jsc,jsc,svga,svgc,sd,lw,s,v,L),Context(svgc,sd,lw,s,L)),Context(svga,fo,f,s,svgc,R)))]
  [],[Context(AdhocContainerPromise),Context(AdhocContainerPromise)]
  [],[Context(AdhocContainerPromise),Context(AdhocContainerPromise)]
```

The test I added is quite inadequate. If you like this but are distressed by the poor testing, I could add some `showcompacts` in the Gadfly tests just to make sure it doesn't crash when given different trees.

PR #149, opened 2015-09-09
Fix Form typealiases

timholy:
On current master, we have
```jl
julia> c = Compose.circle(rand(5), rand(5), rand(5));

julia> isa(c, Compose.Circle)
false

julia> typeof(c)
Compose.Form{Compose.CirclePrimitive{Compose.Measure{Float64,Compose.MeasureNil},Compose.Measure{Compose.MeasureNil,Float64},Compose.Measure{Float64,Compose.MeasureNil}}}
```
This means you can't dispatch on a `Circle`.

This PR fixes that problem.

If I had my druthers, I'd ask for #148 and/or #145 to be merged first; all three of these PRs add a new file `test/misc.jl`, and the other two will be more annoying to rebase.

### DIFF
--- a/src/form.jl
+++ b/src/form.jl
@@ -43,7 +43,7 @@ immutable SimplePolygonPrimitive{P <: Point} <: FormPrimitive
     points::Vector{P}
 end
 
-typealias SimplePolygon Form{SimplePolygonPrimitive}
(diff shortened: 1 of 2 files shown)


timholy:
Thanks for all these merges!

PR #152, opened 2015-09-12
Preserve transparency in fill and stroke

timholy:
I was finding that in Gadfly, `Theme(background_color=c)` wasn't preserving any transparency information in `c`. The root cause was that `color(c)` now strips any transparency (because `Color` does not have transparency), so we need to use a different conversion function if we want to preserve transparency.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -45,6 +45,12 @@ function isinstalled(pkg, ge=v"0.0.0")
 end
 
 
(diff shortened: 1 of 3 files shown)


timholy:
OK, fixed up so that the Gadfly tests pass locally for me with this and https://github.com/dcjones/Gadfly.jl/pull/689 both applied. (Tests will be expected to fail on this until that PR gets merged.)

PR #157, opened 2015-09-22
fix for Union() -> Union{}; require -> using

rened:
(accidentally amended to your commit)


dcjones:
I merged #158 which does that same thing before I saw yours, but thanks for doing this anyway.

Issue #87, opened 2014-10-10
Turn off fontspec for PGF backend


msimberg:
It would be great if it were possible to force the PGF backend to ignore any fonts in the theme so that the plot would just use whatever fonts are used in the latex document.

It seems like this would just be a matter of forcing `img.fontfamily` in the PGF constructor to be `nothing` if a flag is set, but I couldn't quite figure out where it gets set to something else. I could have a try at implementing this myself with some help.

@darwindarak?

bastinat0r:
@msimberg Same here. I want to use a plot created with Gadfly in my paper and the thing that does not work at the moment is removing fontspec. (Of coure a solution would be to just remove all lines with \fontspec)
Anyway I would be glad to know how you solved this

msimberg:
@bastinat0r Unfortunately I don't have a fix for this except for simply removing the lines with fontspec (at least they're easy to remove). I ended up just removing them with my Makefile when building my thesis, which was a fairly pain-free way to do it.

bastinat0r:
Been there done that ;)
For anyone else searching: ``` sed -i '/^\\fontspec/d' $file ```

darwindarak:
`fontspec` is only used to make the outputs consistent across the different backends.  Maybe we can add a keyword argument like `usedefaultfonts` in `draw`.

bastinat0r:
Thats exactly what I had looked for. I thought of calling something like ```draw(PGF(filename, 10cm, 10cm, enablefontspec=false), plot)``` – I don't really know what would be a good name for this parameter.

darwindarak:
I've submitted PR #159 to do this.  Can you guys check to see if this is what you are looking for?

dcjones:
Fixed by Darwin in #159.

PR #159, opened 2015-10-03
Allow use of default TeX fonts in PGF

darwindarak:
This adds a `texfonts` keyword argument to `PGF` that is `false` by default.  It suppresses calls to `fontspec` (see #87).

### DIFF
--- a/src/pgf_backend.jl
+++ b/src/pgf_backend.jl
@@ -74,11 +74,14 @@ type PGF <: Backend
     # Emit only the tikzpicture environment
     only_tikz :: Bool
 
+    # Use default TeX fonts instead of fonts specified by the theme.
+    texfonts::Bool
     function PGF(out::IO,
                  width,
                  height,
                  emit_on_finish::Bool=true,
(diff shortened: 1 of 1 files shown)


bastinat0r:
I have not tested it yet, but from reading through the diff I would say thats exactly what we LaTeX users need. Thanks :)

dcjones:
Looks like a good idea, thanks!

src/pgf_backend.jl
```diff
@@ -74,11 +74,14 @@ type PGF <: Backend
...
 
+    # Use default TeX fonts instead of fonts specified by the theme.
+    texfonts::Bool
     function PGF(out::IO,
                  width,
                  height,
                  emit_on_finish::Bool=true,
-                 only_tikz = false)
+                 only_tikz = false;
+                 texfonts = false;)
```
Keno: Two semicolons.

dcjones: Thanks, fixed.

PR #162, opened 2015-10-06
Fix `hstack` with mixed measures

Keno:
The way this works is that hstack now always results in a cw=1 width
canvas, where any subcanvas with absolute units retains its size and
to make room for that, the remaining canvases are scaled down
proportionately.

### DIFF
--- a/src/stack.jl
+++ b/src/stack.jl
@@ -20,9 +20,12 @@ function hstack(x0, y0, height, aligned_contexts::(@compat Tuple{Context, VAlign
     # fits.
     total_width_units = sum(Float64[context.box.width.cw
                                     for (context, _) in aligned_contexts])
+    total_width_mm = sum(Float64[context.box.width.abs
+                                    for (context, _) in aligned_contexts])
     width = sum(Measure[context.box.width
                         for (context, _) in aligned_contexts])
     width = Measure(width,
+                    abs=total_width_units == 0.0 ? total_width_mm : 0.0,
                     cw=total_width_units > 0.0 ?
                         width.cw / total_width_units : 0.0)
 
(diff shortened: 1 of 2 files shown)


Keno:
Modulo plot styles (note the cut off is in the original plot due to this bug)
screen shot 2015-10-06 at 1 12 52 am
After:
screen shot 2015-10-06 at 1 12 59 am

src/stack.jl
```diff
@@ -20,9 +20,12 @@ function hstack(x0, y0, height, aligned_contexts::(@compat Tuple{Context, VAlign
     # fits.
     total_width_units = sum(Float64[context.box.width.cw
                                     for (context, _) in aligned_contexts])
+    total_width_mm = sum(Float64[context.box.width.cw
```
Keno: this should be .mm of course! I need to go to sleep.

src/stack.jl
```diff
@@ -38,7 +41,8 @@ function hstack(x0, y0, height, aligned_contexts::(@compat Tuple{Context, VAlign
             context.box =
                 BoundingBox(context.box,
                     width=Measure(context.box.width,
-                        cw=context.box.width.cw / total_width_units))
+                        cw=context.box.width.cw / total_width_units,
+                        abs = -total_width_units* context.box.width.cw / total_width_units))
```
Keno: and the first of these should be total_width_mm. Ugh. Fix coming.

Keno:
Another fun example:
```
triangle = compose(context(0.0,0.0,5mm,5mm),polygon([(0, 0), (0.5, 1), (1, 0)]))
compose(context(),hstack(
           compose(context(0.0,0.0,5mm,5mm,rotation=Rotation(pi/2)),triangle),
       compose(context(),line([(0.0,0.5),(1.0,0.5)]),linewidth(2mm)),
           compose(context(0.0,0.0,5mm,5mm,rotation=Rotation(-pi/2)),triangle)),stroke("white"))
```

screen shot 2015-10-06 at 1 48 18 am

On that note is there a way to rotate a context without having to repeat the context size? Or is that just not compose style and the original `triangle` shouldn't have had the triangle dimensions?

tlnagy:
Please open an issue if this is still a problem with `Compose` master. This PR is quite stale now so I'm going to go ahead and close.

PR #163, opened 2015-10-20
WIP: Primitives overhaul

shashi:
This PR removes `Form` and `Property` container types and makes the primitives more first class. `compose` will now take a `FormPrimtive` or `PropertyPrimitive` or equal length vectors of them. The `*Primitive` suffixes can be dropped.

As an example a `Circle` can now be just a center and a radius and not a container of a vector of one `CirclePrimitive`.

Benefits include:
- can create simple list comprehensions for forms and properties (`compose(context(), [... for ... in ...])`)
- clear distinction between scalar and vector properties without the ambiguous `isscalar` checks - this seemed required to set up input from an array of forms in Escher. This change leads to less leak of internal structure outside Compose
- 27 deleted typealiases ;)

TODO
- [x] Remove *Primitive suffixes
- [ ] Bring back `.tag`. I'm planning to have:
  - `settag!(form::Union{AbstractArray, FormPrimitive}, symbol)` which will add the tag to a `ObjectIdDict` and return the form
  - `gettag(form, default)` will give back that symbol
    cc. @timholy what do you think?
- [x] Patchwork back-end changes
- [ ] More tests
- [ ] `writemime` for FormPrimitive, Array{P<:FormPrimitive} (makes it easy to get started)


src/cairo_backends.jl
```diff
@@ -1,4 +1,5 @@
 using Compat
+using Debug
```
rohitvarkey: Is this required?

shashi: of course not. will remove, thanks!

tlnagy:
Any progress on this?

tlnagy:
This PR is very stale at this point. If you get a chance to update, please open a new PR.

PR #164, opened 2015-10-24
Fix breakage on 0.5-dev because of the change in how `a in b` is parsed

yuyichao:
I guess I could figure out which commit/PR introduces this change and enable this only on the necessary versions but this should be good enough and it shouldn't be performance critical anyway....

(Expression patter matching could probably help with this...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,18 @@ language: julia
 os:
     - osx
     - linux
+sudo: required
+dist: trusty
 julia:
     - release
     - nightly
 notifications:
     email: false
+before_install:
+    - if [ `uname` = "Linux" ]; then
+        sudo apt-get update -qq -y;
+        sudo apt-get install libcairo2 libfontconfig1 libpango1.0-0 libpng12-0 libpixman-1-0 gettext;
+      fi
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
     - julia -e 'Pkg.clone(pwd()); Pkg.build("Compose"); Pkg.test("Compose"; coverage=true)'
--- a/examples/sierpinski.jl
+++ b/examples/sierpinski.jl
@@ -1,6 +1,6 @@
 using Compose, Colors
 
-const padding = 0.5mm
+# const padding = 0.5mm
 
 function sierpinski(n::Int)
     if n == 0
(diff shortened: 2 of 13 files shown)


yuyichao:
OTOH, this assertion is really awesome. It would have taken me much longer to figure out the problem without it.

yuyichao:
Is travis enabled for this repo?

yuyichao:
For the reference, the PR that is causing this is https://github.com/JuliaLang/julia/pull/13078

yuyichao:
Seems that a lot of the tests are failing due to the Measures change. I updated the package and the test to fix most of them. However, the comparison of the SVG generated in the test is still failing on my machine. Most of the difference looks like a difference in the id format. However, the `sierpinski.svg` generated in my test seems broken.

yuyichao:
I partially revert one of the recent change on `example/sierpinski.jl` and that seems to fix the svg generated. The order 11 seems too high for the line width and the new `sierpinski` function in it seems to be drawing the triangle in the wrong direction.

The only test failure I get locally are due to id mismatch (example attached). I didn't update the cached svg files since this might be caused by cairo version difference (this is cairo `1.14.2`) (I have no idea who generates the format of these ids...) and you might still want the test to pass on an older version.

```diff
--- test/linecaps.svg   2015-10-25 10:41:53.376712809 -0400
+++ test/data/linecaps.svg      2015-10-25 10:11:40.062456939 -0400
@@ -8,11 +8,11 @@
      stroke-width="0.3"
      font-size="3.88"
 >
-<g stroke-width="2" stroke="#000000" stroke-linecap="butt" id="e6c61b5b-1">
+<g stroke-width="2" stroke="#000000" stroke-linecap="butt" id="fig-e6c61b5b8b554cb79b7700d905a3c167-element-1">
   <path fill="none" d="M5,5 L 15 5"/>
-  <g stroke-linecap="square" id="e6c61b5b-2">
+  <g stroke-linecap="square" id="fig-e6c61b5b8b554cb79b7700d905a3c167-element-2">
     <path fill="none" d="M5,10 L 15 10"/>
-    <g stroke-linecap="round" id="e6c61b5b-3">
+    <g stroke-linecap="round" id="fig-e6c61b5b8b554cb79b7700d905a3c167-element-3">
       <path fill="none" d="M5,15 L 15 15"/>
     </g>
   </g>
```

yuyichao:
OT: One small suggestion, I think the diff is slightly easier to read with `diff -u`.

yuyichao:
I noticed that there's exactly the same test error so I updated the svg files according to my local result and the tests all passes not on [Travis OSX](https://travis-ci.org/yuyichao/Compose.jl/builds/87319455). The remaining test failure on Travis Linux seems to be caused by Cairo version mismatch.

yuyichao:
Yay. Travis tests all pass now by switching to travis `trusty` instead of default one. Strangely, `Cairo.jl` failed to use `apt-get` automatically but installing the package manually before that makes everything happy.

C.C. @dcjones This PR should be fully ready now. You can even enable travis on this repo for regular testing now.

yuyichao:
Note that I think the Cairo versions are still different (1.13.0 vs 1.14.*), but it seems to be new enough to generate otherwise identical output PDF.

yuyichao:
Ref https://github.com/JuliaLang/BinDeps.jl/issues/183 for the Travis installation issue.

dcjones:
Thanks!

PR #167, opened 2015-11-07
Add Immerse backend and tests

timholy:
This adds an official backend for Immerse. This eliminates the need to copy `drawpart` into Immerse, see https://github.com/JuliaGraphics/Immerse.jl/pull/47. It also adds tests for at least some of the functionality that Immerse depends on; it probably isn't comprehensive, but it's a step in the right direction.

If you like some of the functionality in the Immerse tests, of course feel free to cop it. (In general you should feel free to steal anything you want from Immerse/src/compose.jl.) But I didn't want to "pollute" Compose with functionality that others won't need, so I put everything in the test script.

Finally, I modified the tests so that Travis doesn't always fail. If you don't like this, I'll revert this part.

I didn't need to change Gadfly at all; the only part I thought about changing were [these lines](https://github.com/dcjones/Gadfly.jl/blob/fcb249691700cdfde56184049060258ca8758c6d/src/coord.jl#L230-L233) because formerly I had been using `aes.xminview` etc for zooming. Now it seems that `xminview` can only expand the view, not contract it. I switched to using `coord.xmin`, and now it's working again, but if you think `xminview` really should specify the view (or is misleadingly named) perhaps there are additional changes that should be made.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -154,6 +154,7 @@ default_fill_color = colorant"black"
 # Use cairo for the PNG, PS, PDF if it's installed.
 if isinstalled("Cairo")
     include("cairo_backends.jl")
+    include("immerse_backend.jl")
 else
     global PNG
     global PS
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -205,6 +205,8 @@ function absolute_native_units{B}(img::Image{B}, u::Float64)
     img.ppmm * u
 end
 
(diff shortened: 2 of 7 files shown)


timholy:
In the absence of feedback or a "hold," this weekend I'll use my superpowers to merge this. It would also be really nice to tag a new version, but I'm a little more nervous about doing that without approval from @dcjones. It would include the Measures transition, which is not a small thing.

PR #168, opened 2015-11-09
Require Fontconfig when using Cairo-based backends

mortenpi:
As was discussed in dcjones/Gadfly.jl#688, you need Fontconfig with Cairo-based backends to have correct text sizes. So this makes Fontconfig mandatory and clarifies the error message you get. I.e. we go from this
```
julia> PNG()
ERROR: Install Cairo.jl to use the PNG backend. You may need to delete /home/morten/.julia/lib/v0.4/Compose.ji afterwards.
```
to this
```
julia> PNG()
ERROR: Cairo and Fontconfig are necessary for the PNG backend. Run:
  Pkg.add("Cairo")
  Pkg.add("Fontconfig")
You also have to delete /home/morten/.julia/lib/v0.4/Compose.ji
and restart your REPL session afterwards.
```

I did add a whole macro to the global scope to generate the error message. I feel that it is not ideal but it makes the code more readable (i.e. when compared to a bunch of string interpolations).

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -154,8 +154,9 @@ default_fill_color = colorant"black"
 # Use cairo for the PNG, PS, PDF if it's installed.
 macro missing_cairo_error(backend)
     msg1 = """
-    Cairo is necessary for the $(backend) backend. Run:
+    Cairo and Fontconfig are necessary for the $(backend) backend. Run:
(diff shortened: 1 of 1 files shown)


timholy:
LGTM. Thanks!

Issue #172, opened 2015-11-17
Inconsistent behavior of circle(xs,ys,rs)


Sacha0:
`circle(xs, ys, rs)` in 55eabfe21ed94c1a8b88babd4abfb444fb695ef6/src/form.jl#L267-L275 passes radius `r in rs` through `size_measure(r)` in calling the underlying CirclePrimitive constructor, by default interpreting the radius in millimeters, whereas all related methods ultimately pass `r` through `x_measure(r)`, by default interpreting the radius in context units; compare the output of `circle(0.5, 0.5, 0.1)` and `circle([0.5], [0.5], [0.1])`.

Formerly `circle(xs, ys, rs)`'s behavior was consistent with that of the related methods and documentation; this changed in ad3e0387e13cd2fd70ebc78b35cf6ca48c33166f and went mainstream with this morning's release.  Is the inconsistent behavior intended?  If not, what is the intended behavior?

If it would be helpful, given clarification of the intended behavior I would be happy to submit a PR.  Thanks!

timholy:
I can't speak for @dcjones intentions, but I suspect that the goal is to allow one to supply the circle's parameters in a wider range of units. But certainly there needs to be internal consistency. A PR fixing this would be great; we'll do our best to evaluate it in @dcjones's absence, and tag a new release.

Sacha0:
#176

PR #176, opened 2015-12-11
make behavior of circle(xs, ys, rs) consistent (#172)

Sacha0:
This tiny pull request addresses #172: It returns the default measure of `rs` in `circle(xs, ys, rs)` to context units while preserving the ability to call `circle(xs, ys, rs)` with mixed-measure `rs`. Thanks, and best!

### DIFF
--- a/src/form.jl
+++ b/src/form.jl
@@ -271,7 +271,7 @@ function circle(xs::AbstractArray, ys::AbstractArray, rs::AbstractArray, tag=emp
     end
 
     return @makeform (x in xs, y in ys, r in rs),
-        CirclePrimitive{Vec2, Measure}((x_measure(x), y_measure(y)), size_measure(r)) tag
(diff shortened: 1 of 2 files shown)


timholy:
This is failing Travis tests on Linux because of an `apt-get` problem (it was pausing asking for interactive input). I've fixed that in #180; if you add a commit to this pull request (i.e., push to your `circlefix` branch), then it should run again.

Along those lines: is it possible to add a test? Preferably something that fails on current master but succeeds with this PR. That will prevent this from breaking so easily in the future.

Sacha0:
@timholy Rebased and added a test. (Is that test along the lines of what you had in mind?) Thanks, and best!

Sacha0:
Travis failure on Julia master seems unrelated?

timholy:
Yes, it was unrelated (a bug in julia itself, which has been fixed but the binaries have not been updated).

Thanks for fixing this!

Sacha0:
Thanks for the review and merge!

PR #179, opened 2015-12-22
add latex escape chars to pgf backend

bastinat0r:
Properly escape all characters that need to be [escaped](http://www.cespedes.org/blog/85/how-to-escape-latex-special-characters) for latex text mode. ```#\$%&amp;\\^_{}~```

This is especially usefull for symbols created with ```aggregate``` like ```:range_mean``` that would lead to latex errors if not treated properly.

I am however not sure how ```&``` should be treated, as other backends need ```&``` to be escaped as ```&amp;```.
Maybe there should be an additional rule that replaces ```&amp;``` with ```&``` before the ```&``` is treated to compensate that.

### DIFF
--- a/src/pgf_backend.jl
+++ b/src/pgf_backend.jl
@@ -72,7 +72,7 @@ type PGF <: Backend
     emit_on_finish::Bool
 
     # Emit only the tikzpicture environment
-    only_tikz :: Bool
+    only_tikz::Bool
 
     # Use default TeX fonts instead of fonts specified by the theme.
     texfonts::Bool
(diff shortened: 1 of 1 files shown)


bastinat0r:
I do believe https://github.com/dcjones/Compose.jl/pull/111 would solve this ampersand problem to the best. (No need to deal with it in the pgf backend).

timholy:
CC @darwindarak.

darwindarak:
Looks good!  Since only one backend is called at a time, I think it's safe to just go ahead to escape `&` to `\&` for the pgf backend.  Changes from #111 won't affect this PR.

PR #183, opened 2016-01-13
RFC: Remove ASCII requirement on strings when possible

darwindarak:
Same as dcjones/Gadfly.jl#788.  Should all the `ASCIIString`'s be replaced with `AbstractString`'s?


timholy:
Seems like it might be better to us `UTF8String`; this is the direction julia-0.5 is going in (where it will be called `String`).

tlnagy:
This was addressed over the years

PR #187, opened 2016-02-09
DPI keyword argument for one more Image constructor 

ktdrv:
When no output buffer or filename arguments are provided.


timholy:
LGTM. Any objections?

timholy:
Thanks!

PR #188, opened 2016-02-19
Add docstrings to methods in form.jl

jarmokivekas:
I ended up writing docstrings for some of the functions in form.jl in an effort to get a better grasp of julia's `@doc` functionality. I thought I would share them, in case this is something you want.

It might have been better to keep it to a single generic docstring per function, instead of spreading them over several methods, as they now show up in funny order in the `help?>` prompt. If you have opinions/best practices/feedback let me know. I could document the user-facing forms over the weekend if that's something useful.

### DIFF
--- a/src/form.jl
+++ b/src/form.jl
@@ -57,7 +57,12 @@ function polygon()
     return Polygon([PolygonPrimitive(Vec[])])
 end
 
+"""
+    polygon(points)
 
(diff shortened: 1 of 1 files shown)


timholy:
LGTM. I seem to remember a recent pull request or issue in base julia that addresses the sorting issue. So I propose leaving it as you've done it, and it will presumably work well starting in julia-0.5.

Thanks for doing this!

PR #189, opened 2016-02-19
fixed cairo line drawing when NaN values are present

tbreloff:
This method never checked for NaN values, and so when generating a PNG, there would be line segments flying off the screen at NaN values.  This PR changes the logic to ensure we only call `line_to` when both this point and the previous point are non-NaN.  Otherwise it calls `move_to`.  A bonus is that there's no need for the `move_to` before the loop.  This example (using Plots and OnlineAI) was a mess before (each segment is separated by a NaN value), but now works as expected:

tmp

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -862,9 +862,15 @@ end
 
 function draw(img::Image, prim::LinePrimitive)
     if length(prim.points) <= 1; return; end
-    move_to(img, prim.points[1])
-    for i in 2:length(prim.points)
-        line_to(img, prim.points[i])
+    prev_ok = false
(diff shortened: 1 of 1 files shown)


timholy:
Relatedly, what about `Inf`? Should we check `isfinite`, or should we let `Inf` values make weird stuff happen?

tbreloff:
I thought about that, and decided it wasn't as clear-cut. I could see wanting an asymptotic line drawn. If you approximate an asymptote with a "really big number" then it'll mess up axis ranges. 

That said... I'm happy to go along with whatever you think is best. Changing to isfinite is fine by me.

timholy:
I think that could be tackled in a separate PR.

PR #194, opened 2016-03-26
Fix #193, issue with makeform macro in v0.5

ScottPJones:
The AST that it was inspecting had changed in v0.5 with recent changes, so that it didn't detect the `in` operators correctly.
I also fixed another issue that caused a warning in v0.5 (readall has been deprecated and renamed readstring)

PR #195, opened 2016-03-26
Fix problem with HorLineRelPathOp vs HorLineAbsPathOp in svg.jl

ScottPJones:
This looks like it was a typo, and there must not be any unit test that detects it.
Julia v0.5 gave a warning about the method being overridden, and after inspecting the code,
I saw that the real problem was that two methods both had `HorLineRelPathOp` in the signature,
but the first one (outputting 'H') should have had `HorLineAbsPathOp`.
This is a good example of why the new v0.5 behavior is better than silently ignoring methods being overridden.

### DIFF
--- a/src/svg.jl
+++ b/src/svg.jl
@@ -1020,7 +1020,7 @@ function svg_print_path_op(io::IO, op::LineRelPathOp)
 end
 
 
-function svg_print_path_op(io::IO, op::HorLineRelPathOp)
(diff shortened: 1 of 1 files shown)


ScottPJones:
Note: this won't pass until #194 is merged, of course, because without #194, `Compose.jl` doesn't work on v0.5.

PR #196, opened 2016-03-26
Fix bug/typo with wrong PathOp for resolve method

ScottPJones:
This is very similar to the other bug in svg.jl that I found, a method was actually overriding an earlier definition, because of the incorrect type being specified on one of the two methods, but this not being detected until v0.5, because of the old policy of silently overriding methods.


ScottPJones:
Fails on v0.5, simply because everything is broken on v0.5, unless #194 is merged.

ScottPJones:
Thanks!

PR #197, opened 2016-03-29
Fix UndefVarError in v0.5

ScottPJones:
This works around an issue where access to functions has changed in v0.5, if they are in a local
function, and the surrounding function has a variable with the same name.


aviks:
This wasn't caught by the tests?

ScottPJones:
No, and unfortunately, I don't know enough about using `Compose.jl` by itself (yet!) to come up with a unit test for it.  I fixed it by code review, after I saw https://github.com/dcjones/Gadfly.jl/issues/815

lobingera:
unfortunately there seems only sparse testing on Compose.

PR #199, opened 2016-04-13
minor fix to make Gadfly work

montyvesselinov:
Please merge. Gadfly fails to work without this change.

### DIFF
--- a/src/property.jl
+++ b/src/property.jl
@@ -473,7 +473,7 @@ end
 # --------
 
 immutable SVGClassPrimitive <: PropertyPrimitive
-    value::ASCIIString
+    value::AbstractString
 end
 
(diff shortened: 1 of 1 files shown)


ScottPJones:
How does Gadfly fail, in that case?  Is the failure only on v0.5 (master)? It generally isn't a good idea to use an abstract type in an immutable type - if the problem is when handling Unicode characters, does it work if you simply change `ASCIIString` to `UTF8String`?

montyvesselinov:
submitted a new pull request to match the comment above

ScottPJones:
In the future, it's not really necessary to close the pull request and open another one, you could simply make the change, squash the commits, and force push to your branch, and the PR will be updated.

montyvesselinov:
will do.

PR #205, opened 2016-05-16
Fix 0.5 deprecations

Keno:
Also split 0.4 and 0.5 precompiles. For now the 0.5 one is just a copy, but I'll regenerate it for 0.5 soon.


timholy:
Did you get the StackOverflowError when running locally on julia-0.5? Or is there something else going on?

Keno:
Compose is currently triggering all sort of memory corruption. One is https://github.com/JuliaLang/julia/pull/16397. The other one I'm still tracking down.

src/property.jl
```diff
@@ -1,7 +1,9 @@
 
 
 abstract PropertyPrimitive
-
+using Compat
+using Compat: String
```
tkelman: should update minimum Compat version

yuyichao:
Many of these seems to be fixed already. The rest (and more) are included in https://github.com/dcjones/Compose.jl/pull/211

ranjanan:
Is this superseded by #211 ?

PR #223, opened 2016-10-05
fix for deprecation warnings on 0.5

tlnagy:
This fixes some calls to `pointer_to_array`, which was deprecated by https://github.com/JuliaLang/julia/pull/16731

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.4
 Colors
-Compat 0.8.0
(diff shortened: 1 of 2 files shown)


REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.4
 Colors
-Compat 0.8.0
+Compat 0.9.0
```
tkelman: 0.8.0 was fine for these actually

Issue #222, opened 2016-10-04
Pkg.test() on windows with Julia-0.5 fails


mdpradeep:
Here's the error:

```
...
...
text.png:
Binary files C:\Users\julia\AppData\Local\JuliaPro-0.5.0.1\pkgs-0.5.0.1\v0.5\Com
pose\test\data\text.png and C:\Users\julia\AppData\Local\JuliaPro-0.5.0.1\pkgs-0
.5.0.1\v0.5\Compose\test\text.png differ



 in error(::String) at .\error.jl:21
 in include_from_node1(::String) at .\loading.jl:488 (repeats 2 times)
 in process_options(::Base.JLOptions) at .\client.jl:262
 in _start() at .\client.jl:318
while loading C:\Users\julia\AppData\Local\JuliaPro-0.5.0.1\pkgs-0.5.0.1\v0.5\Co
mpose\test\compare_examples.jl, in expression starting on line 37
while loading C:\Users\julia\AppData\Local\JuliaPro-0.5.0.1\pkgs-0.5.0.1\v0.5\Co
mpose\test\runtests.jl, in expression starting on line 28
===============================[ ERROR: Compose ]===============================


failed process: Process(`'C:\Users\julia\AppData\Local\JuliaPro-0.5.0.1\Julia-0.
5.0\bin\julia' -Cx86-64 '-JC:\Users\julia\AppData\Local\JuliaPro-0.5.0.1\Julia-0
.5.0\lib\julia\sys.dll' --compile=yes --depwarn=yes --check-bounds=yes --code-co
verage=none --color=yes --compilecache=yes 'C:\Users\julia\AppData\Local\JuliaPr
o-0.5.0.1\pkgs-0.5.0.1\v0.5\Compose\test\runtests.jl'`, ProcessExited(1)) [1]

================================================================================

INFO: Removing Cairo v0.2.35
INFO: Removing Graphics v0.1.3
ERROR: Compose had test errors
 in #test#61(::Bool, ::Function, ::Array{AbstractString,1}) at .\pkg\entry.jl:74
0
 in (::Base.Pkg.Entry.#kw##test)(::Array{Any,1}, ::Base.Pkg.Entry.#test, ::Array
{AbstractString,1}) at .\<missing>:0
 in (::Base.Pkg.Dir.##2#3{Array{Any,1},Base.Pkg.Entry.#test,Tuple{Array{Abstract
String,1}}})() at .\pkg\dir.jl:31
[...]

PR #225, opened 2016-10-19
remove text example from tests (2nd option for fixing #222)

shashi:


shashi:
what should I do about the nightly test fails? @tkelman

test/compare_examples.jl
```diff
@@ -2,7 +2,20 @@
 cachedout = joinpath(testdir, "data")
 differentfiles = AbstractString[]
 const creator_producer = r"(Creator|Producer)"
-for output in readdir(cachedout)
+
+const KERNEL = VERSION < v"0.5.0-dev" ? OS_NAME : Sys.KERNEL
```
tkelman: use Compat, this cutoff isn't right

tkelman:
looks like https://github.com/JuliaLang/julia/pull/18839 may have broken precompilation of Iterators.jl. I suspected we shouldn't export the Base.Iterators module. Reduce and report to base, or see if removing `Iterators` from `base/exports.jl` is enough to fix it locally

tlnagy:
Any progress on this? Gadfly is also having a lot of warnings due to JuliaLang/julia#18839 and I'm not sure how to patch everything given the reliance on both Iterators and Base.Iterators.

PR #239, opened 2017-02-23
WIP: fixes for deprecations and change in julia 0.6

tlnagy:
This prelim work getting the Gadfly stack ready for Julia 0.6.

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -7,14 +7,14 @@ using Cairo: CairoContext, CairoSurface, CairoARGBSurface,
              CairoEPSSurface, CairoPDFSurface, CairoSVGSurface,
              CairoImageSurface
 
-abstract ImageBackend
-abstract PNGBackend <: ImageBackend
+@compat abstract type ImageBackend end
+@compat abstract type PNGBackend <: ImageBackend end
 
-abstract VectorImageBackend <: ImageBackend
-abstract SVGBackend <: VectorImageBackend
-abstract PDFBackend <: VectorImageBackend
-abstract PSBackend  <: VectorImageBackend
-abstract CairoBackend <: VectorImageBackend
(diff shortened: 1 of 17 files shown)


bjarthur:
what is left to be done here?  if i checkout this PR, i can `using Compose` in the latest julia nightly without any warnings.

tlnagy:
I think I wasn't sure whether all the changes were backwards compatible or not with 0.5. I haven't had time to thoroughly test this PR on both julia 0.5 and 0.6

src/cairo_backends.jl
```diff
@@ -89,7 +89,7 @@ type Image{B <: ImageBackend} <: Backend
     last_ctrl1_point::Nullable{AbsoluteVec2}
     last_ctrl2_point::Nullable{AbsoluteVec2}
 
-    function Image(surface::CairoSurface, ctx::CairoContext, out::IO)
+    @compat function Image{B}(surface::CairoSurface, ctx::CairoContext, out::IO) where {B}
```
bjarthur: what is `B` ?  it doesn't seem to be used at all in the function body or it's input argument definitions.

bjarthur: nevermind...  didn't realize this was a constructor when just looking at the diff.

tlnagy: Yeah the type definition for `Image` looks like this `type Image{B <: ImageBackend} <: Backend`, which is why I had to add the `where {B}` for v0.6 since it doesn't make the implicit assumption for `B` like v0.5 did.

tlnagy:
I restarted the build so lets see if it passes now.

tlnagy:
I think I got stuck and then asked this question: https://discourse.julialang.org/t/trouble-with-types-when-porting-old-compose-jl-code-to-v0-6/2263, I think this is why this PR is still in limbo.

PR #241, opened 2017-04-07
WIP: remove more julia 0.6 dep warns

bjarthur:
they're might be more so don't merge yet

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -313,17 +313,17 @@ end
 
 
 @compat function show(io::IO, ::MIME"image/png", img::PNG)
-    write(io, takebuf_string(img.out))
+    write(io, String(take!(img.out)))
 end
 
 
 @compat function show(io::IO, ::MIME"application/pdf", img::PDF)
-    write(io, takebuf_string(img.out))
+    write(io, String(take!(img.out)))
 end
 
 
 @compat function show(io::IO, ::MIME"application/postscript", img::PS)
-    write(io, takebuf_string(img.out))
+    write(io, String(take!(img.out)))
 end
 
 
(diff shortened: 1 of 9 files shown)


ararslan:
Looks like you got all of the warnings that would be caught by the tests. 👍

bjarthur:
yea, but, when Gadfly's tests are run, more Compose dep warns appear.  so Compose's tests are not complete.  and i can't yet get all of Gadfly's tests to finish yet without errors.

ararslan:
Ah, I see. Thanks for working on this. I'm looking forward to using Gadfly on 0.6. 🙂

bjarthur:
all Gadfly tests pass locally now on julia 0.6 with this Compose PR.  for the Gadfly travis tests to pass on nightly, we're going to need to tag a new release of Compose i think.  since version 0.4.5 we've:

1. stripped out old precompilation code
2. dropped support for 0.4 and added same for 0.6
3. upgraded to new documentation system

shall i merge this PR and tag a release?  any thoughts on a new version number?

ararslan:
+1 for merging and tagging. Since support for Julia 0.4 has been dropped since the last tag, the next tag will need to bump the minor version. That is, since the current tag is v0.4.5, the next should be v0.5.0 per METADATA (and SemVer) requirements.

tlnagy:
You'll need to do

```
PkgDev.tag("Compose", :minor)
PkgDev.publish("Compose")
```

as @ararslan points out

ararslan:
...or use [attobot](https://github.com/attobot/attobot), if it's set up for this organization. 😉

tlnagy:
Good point, I just installed attobot for GiovineItalia

PR #243, opened 2017-04-12
misc fixes for 32bit systems

### LINKED ISSUE #242: Problem with round in misc.jl on 32bit system
The a*b.value at the end of the file seems to exceed the Int32 type in some cases.

    *(a::AbstractFloat, b::Dates.Millisecond) = Dates.Millisecond(round(Integer, (a * b.value)))

After changing the type Integer to Int64, this problem disappered for me.

tlnagy:
closes #242 and GiovineItalia/Gadfly.jl#783

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -1,19 +1,3 @@
-
-
-#function all_equal(xs::AbstractArray)
-    #if isempty(xs)
-        #return true
-    #end
-    #x = xs[1]
-    #for i in 2:length(xs)
-        #if xs[i] != x
-            #return false
-        #end
-    #end
-    #return x
-#end
-
-
 function iszero{T}(x::T)
     return x == zero(T)
 end
@@ -202,7 +186,7 @@ if !method_exists(/, (Dates.Day, Dates.Day))
 end
 
 if !method_exists(/, (Dates.Day, Real))
-    /(a::Dates.Day, b::Real) = Dates.Day(round(Integer, (a.value / b)))
+    /(a::Dates.Day, b::Real) = Dates.Day(round(Int64, (a.value / b)))
 end
 /(a::Dates.Day, b::AbstractFloat) = convert(Dates.Millisecond, a) / b
 
@@ -211,9 +195,9 @@ if !method_exists(/, (Dates.Millisecond, Dates.Millisecond))
 end
 
 if !method_exists(/, (Dates.Millisecond, Real))
-    /(a::Dates.Millisecond, b::Real) = Dates.Millisecond(round(Integer, (a.value / b)))
+    /(a::Dates.Millisecond, b::Real) = Dates.Millisecond(round(Int64, (a.value / b)))
 end
-/(a::Dates.Millisecond, b::AbstractFloat) = Dates.Millisecond(round(Integer, (a.value / b)))
+/(a::Dates.Millisecond, b::AbstractFloat) = Dates.Millisecond(round(Int64, (a.value / b)))
 
 
 if !method_exists(-, (Dates.Date, Dates.DateTime))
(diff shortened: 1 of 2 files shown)


tlnagy:
Hey @trohwer, can you navigate to the Compose directory and run the following commands from terminal inside of the Compose directory:

```
git stash
git checkout -b tlnagy-misc-32bit-fixes master
git pull https://github.com/tlnagy/Compose.jl.git misc-32bit-fixes
```
and see if this fixes the problems you're seeing in #242? I don't have access to a 32bit system so I'll need your help here.

trohwer:
I will do a full check out tomorrow with better internet connectivity. I did a quick check by directly replacing the changed two files in my checkout. This is working for me.

trohwer:
Replacing the Compose project completely by your branch "tlnagy-misc-32bit-fixes master" still yields errors, i.e. from this code in batch.jl:

    grouped_forms = Dict{UInt64, Vector{Form}}()
    grouped_properties = Dict{UInt64, Vector{Property}}()
    for i in 1:n
        h = UInt64(0)
        for property in properties
            h = hash(property.primitives[i], h)
        end

ERROR: MethodError: no method matching hash(::Compose.SVGClassPrimitive, ::UInt64)
Closest candidates are:
  hash(::Compose.SVGClassPrimitive, !Matched::UInt32) at /home/tr/.julia/v0.5/Compose/src/property.jl:497
  hash(!Matched::Gadfly.Col.GroupedColumn, ::UInt64) at /home/tr/.julia/v0.5/Gadfly/src/mapping.jl:14
  hash(!Matched::ForwardDiff.Partials{N,T}, ::UInt64) at /home/tr/.julia/v0.5/ForwardDiff/src/partials.jl:50
  ...
 in optimize_batching(::Compose.Context) at /home/tr/.julia/v0.5/Compose/src/batch.jl:210

I fixed that in my own source already, I think.

Can I generate a patch based on your branch and submit it as a PR?

trohwer:
I create pull request #245 for the remaining instance.

src/misc.jl
```diff
@@ -202,7 +186,7 @@ if !method_exists(/, (Dates.Day, Dates.Day))
 end
 
 if !method_exists(/, (Dates.Day, Real))
-    /(a::Dates.Day, b::Real) = Dates.Day(round(Integer, (a.value / b)))
+    /(a::Dates.Day, b::Real) = Dates.Day(round(Int64, (a.value / b)))
```
tkelman: Are these really needed? This is type piracy and doesn't belong here.

tlnagy: Could you explain what you mean a bit more? Especially since https://github.com/GiovineItalia/Compose.jl/pull/246 is doing something similar

tkelman: If the function is imported from Base, and all input types are from Base, then this method extension has global side effects and can change the behavior of unrelated code depending on whether or not Compose is imported. Unless you're using package local types, this kind of method belongs in the Dates module (or is intentionally not defined there!), it's not Compose's place to be defining it

tlnagy: That makes sense, but many of these functions seemed to be hacks to get Compose working with Dates:

```julia
# Hacks to make Dates time work as coordinates

if !method_exists(/, (Dates.Day, Dates.Day))
    /(a::Dates.Day, b::Dates.Day) = a.value / b.value
end

if !method_exists(/, (Dates.Day, Real))
    /(a::Dates.Day, b::Real) = Dates.Day(round(Int64, (a.value / b)))
end
/(a::Dates.Day, b::AbstractFloat) = convert(Dates.Millisecond, a) / b

if !method_exists(/, (Dates.Millisecond, Dates.Millisecond))
    /(a::Dates.Millisecond, b::Dates.Millisecond) = a.value / b.value
end

if !method_exists(/, (Dates.Millisecond, Real))
    /(a::Dates.Millisecond, b::Real) = Dates.Millisecond(round(Int64, (a.value / b)))
end
/(a::Dates.Millisecond, b::AbstractFloat) = Dates.Millisecond(round(Int64, (a.value / b)))


if !method_exists(-, (Dates.Date, Dates.DateTime))
    -(a::Dates.Date, b::Dates.DateTime) = convert(Dates.DateTime, a) - b
end

+(a::Dates.Date, b::Dates.Millisecond) = convert(Dates.DateTime, a) + b

if !method_exists(-, (Dates.DateTime, Dates.Date))
    -(a::Dates.DateTime, b::Dates.Date) = a - convert(Dates.DateTime, b)
end


if !method_exists(/, (Dates.Day, Dates.Millisecond))
    /(a::Dates.Day, b::Dates.Millisecond) = convert(Dates.Millisecond, a) / b
end

for T in [Dates.Hour, Dates.Minute, Dates.Second, Dates.Millisecond]
    if !method_exists(-, (Dates.Date, T))
        @eval begin
            -(a::Dates.Date, b::$(T)) = convert(Dates.DateTime, a) - b
        end
    end
end


*(a::AbstractFloat, b::Dates.Day) = Dates.Day(round(Int64, (a * b.value)))
*(a::Dates.Day, b::AbstractFloat) = b * a
*(a::AbstractFloat, b::Dates.Millisecond) = Dates.Millisecond(round(Int64, (a * b.value)))
*(a::Dates.Millisecond, b::AbstractFloat) = b * a
```

I don't see how these would be generally useful for users of the Datetime package (since they are pretty plotting specific) so I feel like it is unlikely that we could get them moved over there.

tkelman: I think they're somewhat nonsensical mathematically, you can do arithmetic with periods but with absolute times it doesn't always make sense. You may be using them for plotting specific purposes, but method tables are shared global mutable state, and it's bad form to change them. Method extension on package-specific types is fine, but that's not what this is.

PR #245, opened 2017-04-13
Fix one more instance of hash with hardcoded UInt64

trohwer:
The standard hash seems to be yield UInt. If used with UInt64, this does not work on a 32 bit platform.

This is another case along "misc fixes for 32bit systems #243".


tlnagy:
Thanks!

PR #246, opened 2017-04-19
New divisor function for Dates

Mattriks:
In Gadfly GiovineItalia/Gadfly.jl#784, using `Coord.cartesian(xmin=Date(2014,03,01), xmax=Date(2016,02,01))` currently throws an error not in `Gadfly`, but in `Compose`. This is because `Compose` is a missing a function that allows `Dates.Millisecond` to be divided by `Dates.Day`. This PR adds that function to `Compose`. I've provided an example in GiovineItalia/Gadfly.jl#784 showing that this works with `Date` or `DateTime` objects.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -215,6 +215,11 @@ if !method_exists(/, (Dates.Day, Dates.Millisecond))
     /(a::Dates.Day, b::Dates.Millisecond) = convert(Dates.Millisecond, a) / b
 end
 
+if !method_exists(/, (Dates.Millisecond, Dates.Day))
+    /(a::Dates.Millisecond, b::Dates.Day) = a / convert(Dates.Millisecond, b) 
+end
+
+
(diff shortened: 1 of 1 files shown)


tlnagy:
As per @tkelman's suggestion [here](https://github.com/GiovineItalia/Compose.jl/pull/243#discussion_r112296860) shouldn't this go in base julia?

Mattriks:
Feel free to open an issue in base Julia for all the `Date` math operations in `Compose`.    In the meantime, I'll just add the missing function locally, as I have shown in GiovineItalia/Gadfly.jl#784.

bjarthur:
i'd suggest merging this for now, but then at some point in the future re-writing the parts of Compose which do math on dates to do so in a "sensical" way.

tlnagy:
Agreed. I'm going to go ahead and merge then.

PR #251, opened 2017-05-11
outer constructors; succint functions

bjarthur:
1. per the julia docs recommendation that "It is considered good form to provide as few inner constructor methods as possible", i've moved almost all of the inner constructors to the outside.

2. while at it, i refactored most one-line functions to use the "assignment form"

similar to https://github.com/GiovineItalia/Gadfly.jl/pull/987

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -27,8 +27,6 @@ export compose, compose!, Context, UnitBox, AbsoluteBoundingBox, Rotation, Mirro
        CAIROSURFACE, introspect, set_default_graphic_size, set_default_jsmode,
        boundingbox, Patchable
 
-
-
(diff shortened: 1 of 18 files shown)


tlnagy:
Cool! 🎉

tkelman:
could you avoid doing "rebase and merge"? it drops the reference to the pr #, and if there's a single commit it's otherwise equivalent to "squash and merge"

bjarthur:
oh, okay, will do.  it didn't occur to me that that PR reference was useful.  my thinking was that if the someone merges their own PR then there's no need to record it as a separate contribution.

tkelman:
if there's no commit number visible in the log then it looks as if you just pushed straight to master. which is fine in some packages but loses the cross reference to any discussion like here

PR #252, opened 2017-05-15
simplify parse_colorant interface

bjarthur:
`parse_colorant_vec` is now just `parse_colorant`.  dispatch handles whether the input is a scalar or vector.


bjarthur:
fixed some test too, in particular relating to `text_extents` and `introspect.svg`.

PR #257, opened 2017-06-18
Make backends callable, equiv to calling draw.

dcjones:
This is a real simple change that I'm finding pretty convenient. Lets you pipe things to a backend like:
```julia
plot(cos, 0, 20) |> SVG("cos.svg", 4inch, 3inch)
```

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -187,6 +187,10 @@ const PDF = Image{PDFBackend}
 const PS  = Image{PSBackend}
 const CAIROSURFACE = Image{CairoBackend}
 
+function (img::Image)(x)
+    draw(img, x)
+end
+
 function canbatch(img::Image)
     for vp in values(img.vector_properties)
         isnull(vp) || return false
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -46,6 +46,10 @@ draw(SVG("tomato.svg", 4cm, 4cm), composition)
 nothing # hide
 ```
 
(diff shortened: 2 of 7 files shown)


bjarthur:
ooh, very convenient indeed.  doc and test?

dcjones:
Oh yeah! :+1:

src/cairo_backends.jl
```diff
@@ -187,6 +187,10 @@ const PDF = Image{PDFBackend}
 const PS  = Image{PSBackend}
 const CAIROSURFACE = Image{CairoBackend}
 
+function (img::Image)(x)
+    draw(img, x)
+end
+
```
bjarthur: another small point, to save a couple lines, could say `(img)(x) = draw(img,x)` instead.  i only bring it up because i saw you use this abbreviated notation in your wide-format Gadfly PR for the additional `evalmapping` methods.

bjarthur:
tests pass on 0.5, but not nightly.  strangely, 0.6 was not tested, despite being listed in travis.yml.  i'd say let's merge this PR, but what's up with 0.6 tests?

bjarthur:
ope, i see now.  0.6 tests were turned on a day after this PR was submitted.  i'll recycle the tests...

bjarthur:
0.6 passes, so am merging.  thanks @dcjones !

tlnagy:
Yeah, we weren't testing 0.6 explicitly so when nightlies were switched over 0.7, we ended up not testing 0.6 at all.

PR #263, opened 2017-07-28
fix Compat.String

bjarthur:
... plus a couple typos in the docstrings

### DIFF
--- a/src/form.jl
+++ b/src/form.jl
@@ -361,7 +361,7 @@ end
 @compat const Text{P<:TextPrimitive} = Form{P}
 
 """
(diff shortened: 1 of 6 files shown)


ViralBShah:
It would be nice if a new version (now that the FemtoCleaner is also fixed) can be tagged. Pulling in Gadfly on 0.6 currently otherwise prints a bunch of ASCIIString deprecations, which are fixed in the PR here.

bjarthur:
tagged.  https://github.com/JuliaLang/METADATA.jl/pull/11040

ViralBShah:
Thank you!

PR #267, opened 2017-09-12
Make scalar properties override vector properties.

dcjones:
This is the right way to fix https://github.com/GiovineItalia/Gadfly.jl/issues/770. 

Simple fix, but took while to track down. Currently behavior when drawing something like this
```julia
compose(context(), fill(["red", "blue"]),
    [context(), fill("green"), circle([0.25, 0.75], [0.5], [0.25])])
```
is that the scalar `fill("green")` doesn't override the vector `fill(["red", "blue"])` and it will draw a red and blue circle. That doesn't really make a lot of sense, and can cause issues with batch optimization which rewrites trees turning vector properties into scalar properties. So with this PR that code draws two green circles.

This changes the drawing behavior, I think in a pretty benign way, but this shouldn't be merged until we know what if anything this breaks.

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -293,6 +293,7 @@ function push_property_frame(img::Image, properties::Vector{Property})
             push!(scalar_properties, property)
             push!(applied_properties, typeof(property))
             frame.has_scalar_properties = true
+            img.vector_properties[typeof(property)] = Nullable{Property}()
         elseif !isscalar(property)
             frame.vector_properties[typeof(property)] = property
             img.vector_properties[typeof(property)] = property
--- a/src/pgf_backend.jl
+++ b/src/pgf_backend.jl
@@ -465,6 +465,7 @@ function push_property_frame(img::PGF, properties::Vector{Property})
             push!(scalar_properties, property)
             push!(applied_properties, typeof(property))
             frame.has_scalar_properties = true
+            img.vector_properties[typeof(property)] = Nullable{Property}()
         else
             frame.vector_properties[typeof(property)] = property
             img.vector_properties[typeof(property)] = property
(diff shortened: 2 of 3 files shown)


bjarthur:
Gadfly tests fail with this PR:

```
julia> using Gadfly, DataArrays, RDatasets

julia> set_default_plot_size(6inch, 3inch)

julia> plot(dataset("lattice", "singer"), x=:VoicePart, y=:Height, Geom.boxplot)
Error showing value of type Gadfly.Plot:
ERROR: MethodError: no method matching ColorTypes.RGBA{Float64}(::ColorTypes.RGBA{Float64}, ::Float64)
Stacktrace:
 [1] print_vector_properties(::Compose.SVG, ::Int64, ::Bool) at /Users/arthurb/.julia/v0.6/Compose/src/svg.jl:694
 [2] draw(::Compose.SVG, ::Compose.FormBatch{Compose.CirclePrimitive{Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}},Measures.Length{:mm,Float64}}}) at /Users/arthurb/.julia/v0.6/Compose/src/svg.jl:1097
 [3] drawpart(::Compose.SVG, ::Compose.Context, ::Compose.IdentityTransform, ::Compose.UnitBox{Float64,Float64,Float64,Float64}, ::Measures.BoundingBox{Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}},Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}}}) at /Users/arthurb/.julia/v0.6/Compose/src/container.jl:440
```

tlnagy:
Closed in favor of #315 since I fix the printing error identified above.

PR #269, opened 2017-11-03
print SVG clippaths in sorted order

bjarthur:
does anyone know why clipPaths appear in a random order in the defs section of SVGs?  it makes regression testing in Gadfly painful.  my guess it that it's because they're stored in a Dict, and for some reason julia is iterating over them differently.  not sure at all why, or why even a Dict is necessary.  this PR fixes it, but in a not very graceful post-hoc way.  am open to suggestions, but will merge at some point if no one has any better solutions.

### DIFF
--- a/src/svg.jl
+++ b/src/svg.jl
@@ -164,7 +164,7 @@ type SVG <: Backend
     vector_properties::Dict{Type, Union{(Void), Property}}
 
     # Clip-paths that need to be defined at the end of the document.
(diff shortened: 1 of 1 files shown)


tlnagy:
Not sure why `clippaths` is stored in a dict, but since we already depend on `DataStructures` we could just change the type to an ordereddict to fix this.

bjarthur:
i'd thought about an `OrderedDict`, but didn't take the time to see if Compose already required `DataStructures`.  thanks!  much better.

PR #279, opened 2018-03-22
[WIP] Fixes for julia 0.7 deprecations

tlnagy:
It looks like upstream packages are finally building on the nightlies so it's time to get Compose working as well.

### DIFF
--- a/src/container.jl
+++ b/src/container.jl
@@ -231,7 +231,7 @@ end
 @compat abstract type ContainerPromise <: Container end
 
 # This information is passed to a container promise at drawtime.
-immutable ParentDrawContext
+struct ParentDrawContext
     t::Transform
     units::UnitBox
     box::Absolute2DBox
@@ -334,7 +334,7 @@ end
 
 register_coords(backend::Backend, box, units, transform, form) = nothing
 
-immutable DrawState
+struct DrawState
     pop_poperty::Bool
     container::Container
     parent_transform::Transform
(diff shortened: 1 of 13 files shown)


tlnagy:
We can still run femtocleaner on this PR, right?. There's still plenty of cruft from previous julia version bumps that it can clean up.

bjarthur:
yes, but, it would caught everything you did so far.  no need to put manual sweat into it.  let's wait to merge this until after we tag a release, just right before dataframes merges.

src/container.jl
```diff
@@ -4,18 +4,18 @@
...
 
 # The basic Container which defines a coordinate transform for its children.
-type Context <: Container
+mutable struct Context <: Container
     # Bounding box relative to the parent's coordinates
     box::BoundingBox
 
     # Context coordinates used for children
-    units::Nullable{UnitBox}
+    @compat units::Union{UnitBox, Nothing}
```
bjarthur: i had to [revert](https://github.com/GiovineItalia/Compose.jl/pull/281/commits/066d8f1d7805ea100befc6e2c7f55908369f293c) this and similar changes as they causes tests to fail.  was this for 0.7?

tlnagy: This change was needed for v0.7, but I hadn't verified that my solution worked on v0.6. See https://github.com/JuliaLang/julia/pull/23642

PR #226, opened 2016-10-19
use Requires

shashi:
this seems to fix #221

PR #282, opened 2018-04-04
Add support for Julia v1.0

tlnagy:
This finishes the `Nullable{T}` ==> `Union{T, Nothing}` transition started in #279. I decided that it wasn't worth using the [`Nullables.jl`](https://github.com/JuliaArchive/Nullables.jl) because it's a dead-end archive package and we might as well transition to the new representation of uninitialized values. 

https://github.com/GiovineItalia/Compose.jl/commit/8dd674d483bef75780f18affeeef5f1795877cde is a temporary hack because [`LOAD_CACHE_PATH` was deleted](https://github.com/JuliaLang/julia/pull/26165) in v0.7. I only did this to get tests to pass locally on both v0.6 and v0.7 and I want to make sure CI passes. I'll revert this commit once an actual solution is found (see https://github.com/JuliaLang/Compat.jl/issues/527) before merging


bjarthur:
do you want to try to finish this before we tag?

tlnagy:
The primary hang up is I'm waiting for https://github.com/JuliaGraphics/Cairo.jl/issues/224 to be resolved. I think we should go ahead and tag. 0.7 support can wait.

tlnagy:
It looks like tests are failing partially due to [`isinstalled`](https://github.com/GiovineItalia/Compose.jl/blob/bda806864f03acd8b0c290b23983297bf87c65ba/src/Compose.jl#L34) breaking on 0.7

On 0.7

```julia
julia> include("src/Compose.jl")

## omitted errors

julia> Compose.isinstalled("Fontconfig")
true
```

On 0.6

```julia
julia> include("src/Compose.jl")
Compose

julia> Compose.isinstalled("Fontconfig")
false
```

`Fontconfig` isn't installed in either environment so something else must've broken on 0.7

vtjnash:
The `isinstalled` logic should be replaced by Requires.jl (#226). Would be great to have this finished!

tlnagy:
On @vtjnash's suggestion I started trying to get `Requires.jl` working, but I ran into some problems with `Pkg.resolve` since `Compose` does not have `Cairo` in its `REQUIRE` file. See https://github.com/MikeInnes/Requires.jl/issues/52 for details.

tlnagy:
Thanks to @RalphAS's help and with another round of cleanup, all tests pass on Julia v0.7 (NOTE: This is not backwards compatible due to the iteration protocol changes!). There are a couple silent deprecations that cause the build to crash on v1.0, which I have not yet fixed. Couple things remaining:

- [ ] Fix silent 0.7 deprecations
- [ ] Fix documentation building on v1.0 (looks like this is still in flux ATM)
- [ ] Tag a release prior to merging this PR

tlnagy:
This PR now successfully passes tests on Julia v1.0

src/cairo_backends.jl
```diff
@@ -460,10 +463,10 @@ apply_property(img::Image, property::SVGAttributePrimitive) = nothing
 # --------------
 
 function current_point(img::Image)
-    x = Array{Float64}(1)
-    y = Array{Float64}(1)
-    ccall((:cairo_get_current_point, Cairo._jl_libcairo), Void,
-          (Ptr{Void}, Ptr{Float64}, Ptr{Float64}), img.ctx.ptr, x, y)
+    x = Array{Float64}(undef, 1)
```
bjarthur: do you know what the thinking is behind julia 1.0 requiring `undef` here?  is there something else that could be supplied as the first argument?

.travis.yml
```diff
@@ -4,19 +4,31 @@ os:
...
-script:
-    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'Pkg.clone(pwd()); Pkg.build("Compose"); Pkg.test("Compose"; coverage=true)'
 after_success:
-    - julia -e 'cd(Pkg.dir("Compose")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
-    - julia -e 'cd(Pkg.dir("Compose")); map(x -> Pkg.add(strip(x)), readlines(open(joinpath("docs", "REQUIRE")))); include(joinpath("docs", "make.jl"))'
+    - |
+        julia -e '
+            using Compose, Pkg
+            pkg"add Colors Documenter"
```
bjarthur: should Cairo be removed from docs/REQUIRE?

src/cairo_backends.jl
```diff
@@ -1,11 +1,8 @@
...
-import Cairo
-
-using Cairo: CairoContext, CairoSurface, CairoARGBSurface,
-             CairoEPSSurface, CairoPDFSurface, CairoSVGSurface,
-             CairoImageSurface
+for name in (:CairoContext, :CairoSurface, :CairoARGBSurface, :CairoEPSSurface,
+             :CairoPDFSurface, :CairoSVGSurface, :CairoImageSurface)
+    val = getfield(Cairo,name)
+    @eval const $name = $val
+end
```
bjarthur: what's the reason for the change here?  does `using Cairo: ...` not work with Requires?

src/Compose.jl
```diff
@@ -146,51 +125,48 @@ default_fill_color = colorant"black"
...
-    global PDF
-
-    PNG(args...) = error(@missing_cairo_error "PNG")
-    PS(args...) = error(@missing_cairo_error "PS")
-    PDF(args...) = error(@missing_cairo_error "PDF")
-end
+#global PDF
+PNG(::Any, args...) = error(@missing_cairo_error "PNG")
+PS(::Any, args...) = error(@missing_cairo_error "PS")
+PDF(::Any, args...) = error(@missing_cairo_error "PDF")
```
bjarthur: could these three lines be simplified to `PNG = error...` ?  none of the input args are actually used.

src/Compose.jl
```diff
@@ -146,51 +125,48 @@ default_fill_color = colorant"black"
...
+
 include("svg.jl")
 include("pgf_backend.jl")
 
 # If available, pango and fontconfig are used to compute text extents and match
 # fonts. Otherwise a simplistic pure-julia fallback is used.
 
-if isinstalled("Fontconfig")
+include("fontfallback.jl")
```
bjarthur: to the extent that fontfallback.jl mimics pango well, we should really hype up the fact that Gadfly can generate SVG with pure julia.  maybe add a comment in the Compose README or docs.

bjarthur:
looks great!  thanks for working on this.

we should probably wait to tag a final julia 0.6 release of compose until we can do so simultaneously for gadfly.

once this merges i'll run the scripts to see what other gadfly dependencies are still not working on julia 0.7.

.travis.yml
bjarthur: loading everything that is there would be one less thing to remember to change if docs/REQUIRE were to change in future.  the code to do that is already in master.

src/cairo_backends.jl
bjarthur: i don't understand why `using Cairo: ...` doesn't work on both 0.6 and 0.6...

REQUIRE
```diff
@@ -1,7 +1,7 @@
 julia 0.6
```
bjarthur: should change this to 0.7 if it's not backwards compatible due to the new iteration protocol

bjarthur:
make sure to close #226 when this is merged

simonschoelly:
This pr is the same code as in the branch `tn/more-v07-fixes`, right? Because when I use this branch and enter 
```
using Compose

composition = compose(compose(context(), rectangle()), fill("tomato"));
show(composition)
```
then I get the following error:
```
ERROR: StackOverflowError:
Stacktrace:
 [1] print(::Base.TTY, ::Context) at ./strings/io.jl:29
 [2] show(::Base.TTY, ::Context) at /home/simon/.julia/packages/Compose/lCyaY/src/container.jl:580
 [3] print(::Base.TTY, ::Context) at ./strings/io.jl:31
 ... (the last 2 lines are repeated 18691 more times)
 [37386] show(::Base.TTY, ::Context) at /home/simon/.julia/packages/Compose/lCyaY/src/container.jl:580
 [37387] show(::Context) at ./show.jl:314
```
Doesn't matter if I do this in the REPL or as a script, doesn't matter if it's version 0.7 or 1.0 and also seems to be the case for for other compositions.

I can also open an issue for that, but I thought this might belong here.

tlnagy:
Ack good catch @simonschoelly. I changed a bunch of the showcompact code and knew it would come back to haunt me. Looks like we need to add another unit test

bjarthur:
```
┌ Warning: `Array{T}(m::Int) where T` is deprecated, use `Array{T}(undef, m)` instead.
│   caller = optimize_batching(::Context) at batch.jl:195
└ @ Compose ~/.julia/dev/Compose/src/batch.jl:195
```

bjarthur:
the `isdefined` [here](https://github.com/GiovineItalia/Compose.jl/pull/282/files#diff-ac890e5dee1b06f3e0261601b09364d9R381) needs fixed.

```
┌ Warning: `isdefined(:symbol)` is deprecated, use `@isdefined symbol` instead
│   caller = ip:0x0
└ @ Core :-1
```

bjarthur:
i get three warnings with fontconfig:

```
julia> using Compose, Cairo, Fontconfig
┌ Warning: Use of final value of loop variable `family` around /Users/arthurb/.julia/dev/Compose/src/pango.jl:46 is deprecated. In the future the variable will be local to the loop instead.
└ @ ~/.julia/dev/Compose/src/pango.jl:46
┌ Warning: Deprecated syntax ``const` declaration on local variable` around /Users/arthurb/.julia/dev/Compose/src/pango.jl:35.
└ @ ~/.julia/dev/Compose/src/pango.jl:35
┌ Warning: Error requiring Fontconfig from Compose:
│ TypeError: in link_fontconfig, in ccall: first argument not a pointer or valid constant expression, expected Ptr, got Tuple{Symbol,String}
│ Stacktrace:
│  [1] link_fontconfig() at /Users/arthurb/.julia/dev/Compose/src/Compose.jl:153
│  [2] top-level scope at none:0
│  [3] eval at ./boot.jl:319 [inlined]
│  [4] eval at /Users/arthurb/.julia/dev/Compose/src/Compose.jl:3 [inlined]
│  [5] (::getfield(Compose, Symbol("##115#121")))() at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:67
│  [6] err(::getfield(Compose, Symbol("##115#121")), ::Module, ::String) at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:38
│  [7] #114 at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:66 [inlined]
│  [8] withpath(::getfield(Compose, Symbol("##114#120")), ::String) at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:28
│  [9] (::getfield(Compose, Symbol("##113#119")))() at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:65
│  [10] #invokelatest#1 at ./essentials.jl:691 [inlined]
│  [11] invokelatest at ./essentials.jl:690 [inlined]
│  [12] #3 at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:19 [inlined]
│  [13] iterate at ./generator.jl:47 [inlined]
│  [14] _collect(::Array{Function,1}, ::Base.Generator{Array{Function,1},getfield(Requires, Symbol("##3#4"))}, ::Base.EltypeUnknown, ::Base.HasShape{1}) at ./array.jl:632
│  [15] map at ./array.jl:561 [inlined]
│  [16] loadpkg(::Base.PkgId) at /Users/arthurb/.julia/packages/Requires/9Jse/src/require.jl:19
│  [17] #invokelatest#1 at ./essentials.jl:691 [inlined]
│  [18] invokelatest at ./essentials.jl:690 [inlined]
│  [19] require(::Base.PkgId) at ./loading.jl:855
│  [20] macro expansion at ./logging.jl:311 [inlined]
│  [21] require(::Module, ::Symbol) at ./loading.jl:834
│  [22] eval(::Module, ::Any) at ./boot.jl:319
│  [23] eval_user_input(::Any, ::REPL.REPLBackend) at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v0.7/REPL/src/REPL.jl:85
│  [24] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v0.7/REPL/src/REPL.jl:117 [inlined]
│  [25] (::getfield(REPL, Symbol("##28#29")){REPL.REPLBackend})() at ./task.jl:262
└ @ Requires ~/.julia/packages/Requires/9Jse/src/require.jl:40
```

src/Compose.jl
```diff
@@ -1,22 +1,20 @@
-VERSION >= v"0.4.0-dev+6521" && __precompile__()
+__precompile__()
```
ararslan: If you require 0.7 at a minimum, you can remove the call to `__precompile__()` altogether, as it's now the default.

README.md
```diff
@@ -1,6 +1,6 @@
 # Compose!
 
-[![][docs-latest-img]][docs-latest-url] [![][pkg-0.6-img]][pkg-0.6-url] [![][pkg-1.0-img]][pkg-1.0-url] [![][travis-img]][travis-url] [![][codecov-img]][codecov-url]
+[![][docs-latest-img]][docs-latest-url] [![][pkg-0.7-img]][pkg-0.7-url] [![][pkg-1.0-img]][pkg-1.0-url] [![][travis-img]][travis-url] [![][codecov-img]][codecov-url]
```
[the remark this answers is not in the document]
tlnagy: Is the recommendation to drop the badges for now?

tlnagy:
Compose's coverage is fairly weak so I suspected that a lot of things slipped through. Going to try and plug all the holes.

bjarthur:
more tests would be great!

tlnagy:
Yeah, we'll definitely need to do more direct testing in this repo instead of relying on Gadfly to catch everything. Any ideas on the Fontconfig error? I have no idea why the `ccall`s are failing, those should work the same as on v0.6

bjarthur:
dunno about the `ccalls`.  the hack is to hard code the string literals by cutting and pasting them from `Cairo/deps/deps.jl`.  but then i get world age warnings when `using Compose, Cairo, Fontconfig`.

a bigger question i have is whether we can disentangle Fontconfig and Pango from Cairo.  i could see how someone might want to use Compose/Gadfly to generate SVGs, which does *not* require Cairo, but would want to use Fontconfig and Pango for text rendering.  do you think that is possible?

lobingera:
@bjarthur I've seen this just by coincidence. Yes, of course you could split Cairo.jl into more specialized packages, but i guess you will save not much. Providing the binary dependencies are the hard part, not how you handle this on julia side with types and ccalls.

src/cairo_backends.jl
```diff
@@ -314,7 +317,7 @@ function pop_property_frame(img::Image)
     frame.has_scalar_properties && restore_property_state(img)
 
     for (propertytype, property) in frame.vector_properties
-        img.vector_properties[propertytype] = Nullable{Property}()
+        img.vector_properties[propertytype] = Union{Property, Nothing}()
```
bjarthur: should be ` = nothing`

README.md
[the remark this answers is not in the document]
ararslan: Up to you, just note that 0.7 will be outdated and 1.0 doesn't exist. There's no official recommendation at the moment as far as I know.

ararslan:
Tests are passing!! :tada: Excellent work here, @tlnagy! I'm really excited to use Gadfly on 1.0.

tlnagy:
Thanks @ararslan, but we still have to figure out the Fontconfig problems that @bjarthur brought up. Broadly, I'm okay with temporarily dropping our Fontconfig support and rely on our `fontfallback.jl` code and then work out a solution for 1.0. We can't wait any longer with this and I don't know of a good solution to this right now.

bjarthur:
so `using Compose, Cairo` works, but `using Compose, Cairo, Fontconfig` does not?  is that what you're saying?

will checkout the last version of this PR tonight and test with Gadfly.

tlnagy:
> so using Compose, Cairo works, but using Compose, Cairo, Fontconfig does not? is that what you're saying?

Exactly and this is due to `ccall` problems with the conditional code that is loaded from `pango.jl` via `Requires`.

bjarthur:
with the following patch to this PR, gadfly CI tests pass using https://github.com/GiovineItalia/Gadfly.jl/pull/1189 and https://github.com/GiovineItalia/Hexagons.jl/pull/10:

```
diff --git a/src/batch.jl b/src/batch.jl
index 4b12130..f1d1c3e 100644
--- a/src/batch.jl
+++ b/src/batch.jl
@@ -192,7 +192,7 @@ function optimize_batching(ctx::Context)
 
         if !haskey(grouped_forms, h)
             grouped_forms[h] = Form[similar(form) for form in forms]
-            group_prop = Array{Property}(length(properties))
+            group_prop = Array{Property}(undef, length(properties))
             for j in 1:length(properties)
                 group_prop[j] = Property([properties[j].primitives[i]])
             end
diff --git a/src/property.jl b/src/property.jl
index 82a9caf..3904a15 100644
--- a/src/property.jl
+++ b/src/property.jl
@@ -434,7 +434,7 @@ function resolve(box::AbsoluteBox, units::UnitBox, t::Transform, primitive::JSCa
     i = 1
     validx = 1
     while true
-        j = findnext(primitive.code, '%', i)
+        j = findnext(isequal('%'), primitive.code, i)
 
         if j === nothing
             write(newcode, primitive.code[i:end])
```

with one exception:

```
[ Info: label_size.pdf
┌ Warning: broadcast will default to iterating over its arguments in the future. Wrap arguments of
│ type `x::Measures.Length{:mm,Float64}` with `Ref(x)` to ensure they broadcast as "scalar" elements.
│   caller = ip:0x0
└ @ Core :-1
```

not sure whether the problem is in compose or gadfly.  hard to track down without a stack trace.


regression tests are mostly fine, with four main problems

1.  gadfly is choosing a different set of ticks for some reason in about a dozen tests.  issue1004.jl for example.

2. histogram bars are not drawn in the same z-order for some reason in about a half dozen tests.  stacked_discrete_histogram.jl for example.  have to zoom in to see it.

3. histogram2d.jl has it's sparse squares shifted around by a small but disturbing amount

4. the histograms in the top two panels in subplot_grid_free_y_2.jl are completely different.


i wouldn't be adverse to merging the three relevant PRs and then advertising a gadfly beta on discourse for field testing before tagging a release for julia 0.7.

tlnagy:
> not sure whether the problem is in compose or gadfly. hard to track down without a stack trace.

Have you tried asking over on discourse? That's basically impossible to track down with so little info.

> i wouldn't be adverse to merging the three relevant PRs and then advertising a gadfly beta on discourse for field testing before tagging a release for julia 0.7.

Agreed. While the regressions are worrisome, they aren't showstoppers. We need to at least have a beta at this point. I'm in favor of merging into master and then letting people take it for a spin.

bjarthur:
the broadcast / Ref() warning is now fixed, and i believe all the regression errors are due to the change in Random.  so...  let's do this!

tlnagy:
I just squashed and rebased. This ready to merge. 

We definitely need to resolve https://github.com/GiovineItalia/Compose.jl/pull/282#issuecomment-416537701 soon, but I'm okay with this getting merged and the fix for that being a separate PR.

tlnagy:
https://github.com/GiovineItalia/Compose.jl/pull/282/commits/b7319746e5bd3f55cde1aba8c791263513280b12 fixes the problem identified by @simonschoelly so now printing works as expected in the noncompact/default case too.

src/cairo_backends.jl
tlnagy: Changing this back leads to warnings (maybe @RalphAS can help us understand why?) 

```julia
julia> using Compose
┌ Warning: Package Compose does not have Cairo in its dependencies:
│ - If you have Compose checked out for development and have
│   added Cairo as a dependency but haven't updated your primary
│ - Otherwise you may need to report an issue with Compose
└ Loading Cairo into Compose from project dependency, future warnings for Compose are suppressed.
```

bjarthur:
@Mattriks care to submit a PR to Measures.jl with your solution?  much better than mine!

src/container.jl
```diff
@@ -69,9 +69,9 @@ mutable struct Context <: Container
 end
 
 Context(x0=0.0w, y0=0.0h, width=1.0w, height=1.0h;
-            units=Nullable{UnitBox}(),
-            rotation=Nullable{Rotation}(),
-            mirror=Nullable{Mirror}(),
+            units=Union{UnitBox, Nothing}(),
+            rotation=Union{Rotation, Nothing}(),
+            mirror=Union{Mirror, Nothing}(),
```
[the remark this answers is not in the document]
bjarthur: @tlnagy did you see this comment?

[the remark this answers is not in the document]
tlnagy: I missed this, doing it right now

tlnagy:
AFAIK, if tests still pass this PR is ready to be merged. I've tried to address all points raised in the review.

tlnagy:
Looks like there are no major objections so I'm going to go ahead and pull the trigger. 

We should let 1.0 support stew on master for a week give or take and then tag a new release. This'll give people time to put it through its paces and make sure that everything is working well.

PR #285, opened 2018-06-01
Drop some IterTools methods for Base.Iterators methods

iamed2:
These two _might_ get deprecated in IterTools so I want to make sure they aren't necessary. 

Note that the ordering for Base.Iterators.product is different than the ordering for IterTools.product, and there are no Compose tests which follow that path.


iamed2:
Looks like Gadfly tests caught that Compose was expecting `product()` to exist

iamed2:
Alright, this should work now.

bjarthur:
there might be Gadfly tests which would test the differing order in `product`.  might try running the [regressions tests](http://gadflyjl.org/latest/dev/regression.html).

iamed2:
Okay will do! The regular Gadfly tests did pass with this change.

iamed2:
Regression tests pass! :)

bjarthur:
thanks!

PR #287, opened 2018-08-07
Strokedash edit

Mattriks:
I ran into this issue while working on Gadfly `Guide.linekey`.
In `Compose`, `strokedash(values::AbstractArray{AbstractArray})` won't work because e.g.:
```julia
Vector{Vector} <: AbstractArray{AbstractArray} # is false
```
The correct syntax is `AbstractArray{<:AbstractArray}`. So with this edit, the following now works:

```julia
using Compose, Gadfly
line_style = [:solid, :dash, :dot, :dashdot, :dashdotdot]
linestyles = Gadfly.get_stroke_vector.(line_style)
lines = [[(0.0,x),(1.0,x)] for x in linspace(0.1,0.9, 5)] 

p = compose(context(),
    (context(), Compose.line(lines), strokedash(linestyles), stroke("deepskyblue"))
)
```
strokedash

### DIFF
--- a/src/property.jl
+++ b/src/property.jl
@@ -86,7 +86,7 @@ end
 const StrokeDash = Property{StrokeDashPrimitive}
 
 strokedash(values::AbstractArray) = StrokeDash([StrokeDashPrimitive(collect(Measure, values))])
-strokedash(values::AbstractArray{AbstractArray}) =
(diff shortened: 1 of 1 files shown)


tlnagy:
Good catch!

PR #288, opened 2018-08-10
implement optional loading of Cairo and Fontconfig

RalphAS:
This is one way to handle the optional loading issue, which would allow you to tag for v0.7. (Eventually the Pkg maintainers may sanction something official, but why wait?)  I didn't intentionally break compatibility with v0.6, but I didn't test for it either.

Note that random numbers are not consistent across Julia versions, so the tests fail as they stand; actual images were indistinguishable to this mere mortal.

You should add v0.7 to the CI matrix; there are lots of deprecations which fail on nightly but might be ok for v0.7.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -124,6 +124,37 @@ default_line_width = 0.3mm
 default_stroke_color = nothing
 default_fill_color = colorant"black"
 
+if VERSION >= v"0.7-"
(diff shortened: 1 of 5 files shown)


RalphAS:
@tlnagy The simpler alternative in #289 may be preferable.

tlnagy:
Closed in favor of #289

PR #289, opened 2018-08-12
Accommodate Requires for Julia v0.7

RalphAS:
This version works if callers of PNG etc. import Cairo themselves.

Tests will fail because random number sequences have changed for v0.7 (but images seem indistinguishable).


tlnagy:
Thanks a lot for this solution. I'm working on tidying up #282 and hopefully we can get it merged soon!

Issue #116, opened 2015-04-17
Working with paths


vchuravy:
So I decided to use `Compose` to computationally create tags. I managed to create a figure that looks correct in `IPython` by using `paths` which seems to be a under documented but very useful feature. But when I try to export this to `PNG` or `PDF` the parts of the figure that are defined via paths disappear and when I use `SVG` the picture looks right but there is a small rendering bug.

The code is a bit involved so I put it in a gist https://gist.github.com/vchuravy/86f1383def946288c7b2

shashi:
I have encountered the rendering bug too. If you take a path from an SVG made in Inkscape and parse it into the compose Path form, it's not going to look the same when I render it back to SVG. The way I got around this is by adding more nodes to the path in Inkscape and trying it out with Compose until I got a satisfactory picture.

As for `PNG` and `PDF` backends, they do not have methods to render paths yet.

dcjones:
@shashi is right, I haven't implemented paths on the cairo backends yet. Can you point out the rendering bug? This is what I get on `SVG`.

![screen shot 2015-04-17 at 10 31 00 am](https://cloud.githubusercontent.com/assets/210361/7207264/ed81ed54-e4ec-11e4-9ec0-e1abf120b2a3.png)

dcjones:
I'm adding support for paths in the cairo backend now. I made some abstract art for us to enjoy in the meantime.

![try](https://cloud.githubusercontent.com/assets/210361/7207996/57282e8a-e4f2-11e4-9571-ee38c9cf24d3.png)

dcjones:
Paths should render correctly on PNG, PDF, PS. I found a couple bugs on SVG that I fixed in that commit as well. If you continue to see rendering bugs, please reopen.

vchuravy:
Sorry for my late reply.
The rendering bug I saw was when opening the file with gimp. Inkscape
managed to handle it properly.

Thank you for the path implementation in the cairo backend. Compose is
quickly shaping up to be one of my favourite tool for creating graphics
besides tikz/pgf

PR #298, opened 2018-09-12
Arc Primitives

### LINKED ISSUE #227: How do I draw an arc?
This is either missing from the documentation or I am unable to find how to do this.  If I want to draw an arc (that represents part of the circumference of a circle), how would I do that with Compose.jl?  Thanks!

Mattriks:
### This PR:
- [x] Adds `Compose.arc/slice` for drawing arcs and pie slices (closes #227)  
* The method is `arc(x, y, r, angle1, angle2, slice::Bool)`.   
* `slice` is `arc` with `slice=true`.   
* Arcs are consistently drawn clockwise from `angle1` to `angle2`.   

- [x] Works for Cairo, SVG & PGF backends
- [x] I've also removed some out-of-date Path Primitive code (see Example 2 below for more info)  
- [x] Basic documentation
 - [x]  Test implemented

### Examples:
```julia
# Example 1

colv = ["red","orange","green","blue", "purple"]
a = range(-π/4, stop=7π/4, length=6)+ 0.2*randn(6)
a[6] = a[1]

# arc/slice can toke scalars or vectors
img1 = compose(context(),
    (context(), slice([0.5], [0.5], [0.3], a[1:5], a[2:6]), fill(colv)) )
img2 = compose(context(),
    (context(), slice([0.5], [0.5], [0.32], a[1:5], a[2:6]), stroke("white"), fill(colv), linewidth(1.4mm)) )
img3 = compose(context(),
    (context(), slice([0.5], [0.5], [0.3], a[1:5], a[2:6]), stroke(colv), fill("transparent"), linewidth(1.4mm)) )
# Default arg slice = false:
img4 = compose(context(),
    (context(), arc([0.5], [0.5], [0.3], a[1:5], a[2:6]), stroke(colv), fill("transparent"), linewidth(2mm)) )

draw(PNG(8inch,2inch), hstack(img1, img2, img3, img4))
```
arcex1

In Example 2, I reproduce the example from the gist in #116. That gist uses old Path Primitive code, which is undocumented, hasn't been maintained, and is not user-friendly (the gist has ~100 lines of user code). 
Here is my user-friendly code:
```julia
# Example 2

a = range(-0.5π, stop=1.5π, length=13)
colv = repeat(["white","black"],outer=6)

 img = compose(context(),
        (context(), slice([0.5], [0.5], [0.4], a[1:12], a[2:13]), fill(colv)),
        (context(order=2), arc([0.5], [0.5], [0.2], [0, π], [π,0]), 
            fill(["black","white"]), stroke(["white","black"]), linewidth(6pt))
    ) 
draw(PNG(3inch,3inch), img)
```
arcex2


tlnagy:
I gave this a quick glance over and it looks great! The biggest thing that I would add are better docstrings for the user facing functions of slice and arc. They should describe the parameters and provide one or two examples of usage.

Mattriks:
I've refactored the Compose docs Gadfly-style, and put examples of `arc` and `slice` in the new Compose gallery. The Compose docs need a bit of work, so I'll leave adding examples of the other forms to a future PR. For this current PR I still have 2 other items on my task list.

src/form.jl
```diff
@@ -575,408 +575,83 @@ boundingbox(form::BitmapPrimitive, linewidth::Measure, font::AbstractString, fon
...
+    slice(x, y, r, θ1, θ2)
 
-resolve(box::AbsoluteBox, units::UnitBox, t::Transform, p::QuadCurveRelPathOp) =
-        QuadCurveRelPathOp(
-            (resolve(box, units, t, p.ctrl1[1]),
-             resolve(box, units, t, p.ctrl1[2])),
-            (resolve(box, units, t, p.to[1]),
-             resolve(box, units, t, p.to[2])))
+Define a pie slice with its center at (`x`,`y`), radius of `r`, between `θ1` and `θ2`  
+"""
```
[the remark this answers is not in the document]
Mattriks: Examples are in the new Compose gallery (see my comment below), rather than in docstrings: same as in Gadfly.

Mattriks:
Done!  `Gadfly.Geom.pie` anyone?

src/abandoned.jl
```diff
@@ -0,0 +1,816 @@
+
```
tlnagy: Is this all stuff that isn't needed any more? Why are we keeping it?

Mattriks: At the moment, I can't think of a reason why you would want Path Primitives, in addition to the Compose forms (line, curve etc). If I don't think of a reason for wanting it, I'll remove the abandoned code in a separate PR.

tlnagy: Yeah, that'll give us time to think of a good way to handle it and we can merge this when you do

tlnagy: Sorry i wasn't clear. I meant that those as two separate things, i.e. handling the abandoned code in separate PR and merging this PR now.

bjarthur:
https://github.com/GiovineItalia/Compose.jl/blob/master/src/cairo_backends.jl#L469 is the only call to libcairo in all of Compose.jl.  could the containing function also be be moved to src/abandoned.jl?

Mattriks:
I agree, `current_point` can be moved to abandoned.jl, and abandoned.jl can be removed. Great to see the Compose code being tidied up.

bjarthur:
`slice` might get confused with `sortslices` and `mapslices`.  might `wedge` be a better name?

Mattriks:
Wedge has different [geometric definition](https://en.wikipedia.org/wiki/Wedge_(geometry)). But I would agree to [sector](https://en.wikipedia.org/wiki/Circular_sector).

tlnagy:
Do we want to make this change before we tag a new release?

Mattriks:
I see now that what is needed a new Compose form e.g. `Compose.beziergon` (or `bezigon`). That should be possible (at least in svg).

PR #300, opened 2018-09-15
Fix method error in table.jl

jaakkor2:
Fixes problem with the following plot on Julia v1.0.0.  Otherwise I get
`ERROR: MethodError: no method matching +(::Array{Float64,1}, ::Float64)`
```
plot(
    layer(x=[-2.5, 2.5, 2.25, 2.5, 2.25],y=[0, 0, 0.15, 0 ,-0.15],
        Geom.line(preserve_order=true)),
        Coord.cartesian(aspect_ratio=MathConstants.golden)
 )
```

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -100,8 +100,8 @@ function force_aspect_ratio!(tbl::Table,
         w = w0
         h = h0 * adj
         delta = h0 - h
-        y_solution[1:tbl.y_focus.stop] += delta/2
-        y_solution[tbl.y_focus.stop+1:end] -= delta/2
+        y_solution[1:tbl.y_focus.stop] .+= delta/2
+        y_solution[tbl.y_focus.stop+1:end] .-= delta/2
         h_solution[tbl.y_focus] *= adj
     end
 end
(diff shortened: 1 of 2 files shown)


tlnagy:
Thanks for taking the time to make a PR. Could you add a minimal test for this?

jaakkor2:
One-liner test in Gadfly would be
```
plot(layer(y=[1,2,3]),Coord.cartesian(aspect_ratio=1.6))
```
Next step is to figure out the test for Compose.

jaakkor2:
Compose specific test line
```
Compose.force_aspect_ratio!(Compose.Table(1, 1, UnitRange(1,1), UnitRange(3:3), aspect_ratio=1.6), [0.0, 5.8, 11.8], [0.0, 85.3], [5.8, 6.0, 119.6], [85.3, 4.7])
```

src/table.jl
```diff
@@ -100,8 +100,8 @@ function force_aspect_ratio!(tbl::Table,
         w = w0
         h = h0 * adj
         delta = h0 - h
-        y_solution[1:tbl.y_focus.stop] += delta/2
-        y_solution[tbl.y_focus.stop+1:end] -= delta/2
+        y_solution[1:tbl.y_focus.stop] .+= delta/2
+        y_solution[tbl.y_focus.stop+1:end] .-= delta/2
         h_solution[tbl.y_focus] *= adj
```
tlnagy: I'm surprised you didn't have to update this line to broadcast as well.

jaakkor2: Not needed, see `@edit [1,2,3]*2`. On line 97 dot is unnecessary.

tlnagy: Huh, seems like a weird exception.

PR #303, opened 2018-09-16
Forms_gallery

Mattriks:
This PR is a doc update. It adds
- [x] a new Forms gallery (check it out!)
- [x] fixes some relevant code in form.jl
- [x] docstrings for some forms


tlnagy:
Thanks for taking the time to do this. Compose never got the love that the Gadfly docs did and I'm glad to see it come together.

src/form.jl
```diff
@@ -557,11 +598,16 @@ function bitmap(mime::AbstractString, data::Vector{UInt8}, x0, y0, width, height
     return Bitmap{typeof(prim)}([prim], tag)
 end
 
+"""
+    bitmap(mimes::AbstractArray, datas::AbstractArray, x0s::AbstractArray, y0s::AbstractArray, widths::AbstractArray, heights::AbstractArray)
```
tlnagy: You sometimes include the types in the doc strings and sometimes don't. I think we should not include it. 

We can also autogenerate them like we do in Gadfly using DocStringExtensions

Mattriks: The form functions e.g. `bitmap`, `curve` etc have two methods: one for scalar arguments and one for Array arguments.  I followed this [example](https://docs.julialang.org/en/v1/manual/documentation/#Functions-and-Methods-1) in the Julia docs.

Mattriks: Yes some docstrings that were already there for the Array methods are missing the Array typing: I should put it in.

Mattriks:
Done. Also rebased and tidied up commits.

PR #305, opened 2018-09-20
re-enable keyword args to PNG, etc.

bjarthur:
`PNG(...; dpi=192)` no longer works after the refactoring for julia 0.7.  this fixes it.


tlnagy:
Good catch. I'm surprised none of our tests caught this. We should definitely test this.

PR #312, opened 2018-09-30
fix conditional dependencies

bjarthur:
don't merge yet.  still need to finish Pango.jl.


bjarthur:
still working on pango, and it's going to be awhile.  i've got  Clang.jl-generated wrappers for all of the libpango header files.  but still trying to use BinaryBuilder.jl to build libpango.  see https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/367 for a list of the dependencies.

PR #313, opened 2018-10-01
allow for direct call for link_cairo

### LINKED ISSUE #308: Package dependent code loading doesn't work in v0.7 ?
I'm trying to update Immerse to v0.7 and I'm getting an issue with this part:

https://github.com/GiovineItalia/Compose.jl/blob/master/src/Compose.jl#L167

This mechanism seems to work fine in the REPL:

```
julia> using Cairo

julia> using Compose

julia> Compose.absolute_native_units
absolute_native_units (generic function with 4 methods)


julia> module A
           using Cairo
           using Compose
           
           import Compose.absolute_native_units            
       end
Main.A
```

But if I put the code above in a package, it fails to import absolute_native_units:

```
julia> using TestCompose
[ Info: Recompiling stale cache file /Users/jbieler/.julia/compiled/v0.7/TestCompose/Tl9GO.ji for TestCompose [c558b20d-ff79-52a3-ae8b-9e51a808e850]
┌ Warning: Package TestCompose does not have Compose in its dependencies:
│ - If you have TestCompose checked out for development and have
│   added Compose as a dependency but haven't updated your primary
│ - Otherwise you may need to report an issue with TestCompose
└ Loading Compose into TestCompose from project dependency, future warnings for TestCompose are suppressed.
WARNING: could not import Compose.absolute_native_units into TestCompose
```

Any way to fix this ? 

Might be related to https://github.com/GiovineItalia/Compose.jl/issues/250, but I'm not sure what's the status of it.

jonathanBieler:
This should fix #308. I'm not totally sure about the import changes, but it seems to work on my computer.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -159,8 +159,8 @@ end
 
 function link_cairo()
     @info "Loading Cairo backend into Compose.jl"
-    include("cairo_backends.jl")
-    include("immerse_backend.jl")
+    include(joinpath(@__DIR__,"cairo_backends.jl"))
+    include(joinpath(@__DIR__,"immerse_backend.jl"))
 end
 
 function __init__()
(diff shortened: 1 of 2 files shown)


bjarthur:
lgtm though i haven't tried it locally.  @tlnagy

PR #314, opened 2018-10-07
Fillopacity

### LINKED ISSUE #181: opacity and alpha in pdf and png output
The current pdf output does not seem to support opacity or alpha values. I'm guessing this is due to the Cairo PDF surface version not being high enough? I didn't see a quick way of fixing this though.

Here's a minimal example

```
using Compose
fig = compose(context(), 
        (context(0.0, 0.0, 0.5, 0.5), circle(), fill("tomato")),
(context(0.5, 0.5, 0.5, 0.5), circle(), fill("tomato"), fillopacity(0.25)))
img = PDF("test-opacity-alpha.pdf",100mm,100mm)
draw(img,fig)
imgpng = PNG("test-opacity-alpha.png",100mm,100mm)
draw(imgpng,fig)
```

The SVG file produced in jupyter looks fine though.

Mattriks:
This PR:
- Fixes #181 (so that `fillopacity` is not order-dependent)
- Enables scalar and vector `fillopacity` and forms to work together
- Begins a new `Properties` gallery for the Compose docs (will be added to other PRs)

Example:
```
set_default_graphic_size(14cm,4cm)
img = compose(context(),
  (context(), circle(0.5, 0.5, 0.08), fill("orange"), fillopacity(0.3)),
  (context(), circle([0.1, 0.26], [0.5], [0.1]), fillopacity(0.3), fill("blue")),
  (context(), circle([0.42, 0.58], [0.5], [0.1]), fill(["yellow","green"]), fillopacity(0.3)),
  (context(), circle([0.74, 0.90], [0.5], [0.1]),  fillopacity([0.5,0.3]), fill(["yellow","red"]))
)

draw(PNG(), img)
# draw(SVG(), img) # will also work
```
iss_314

### DIFF
--- a/src/property.jl
+++ b/src/property.jl
@@ -69,7 +69,19 @@ end
 const Fill = Property{FillPrimitive}
 
 fill(c::Nothing) = Fill([FillPrimitive(RGBA{Float64}(0.0, 0.0, 0.0, 0.0))])
+
+"""
+    fill(c)
+
+Define a fill color, where `c` can be a `Colorant` or `String`.
+"""
 fill(c::Union{Colorant, AbstractString}) = Fill([FillPrimitive(parse_colorant(c))])
+
+"""
+    fill(cs::AbstractArray)
+
+Arguments can be passed in arrays in order to perform multiple drawing operations at once.
+"""
 fill(cs::AbstractArray) = Fill([FillPrimitive(c == nothing ?
         RGBA{Float64}(0.0, 0.0, 0.0, 0.0) : parse_colorant(c)) for c in cs])
 
@@ -196,7 +208,18 @@ end
 
 const FillOpacity = Property{FillOpacityPrimitive}
 
+"""
+    fillopacity(value)
+
+Define a fill opacity, where 0≤value≤1.
+"""
 fillopacity(value::Float64) = FillOpacity([FillOpacityPrimitive(value)])
+
+"""
+    fillopacity(values::AbstractArray)
+
+Arguments can be passed in arrays in order to perform multiple drawing operations at once.
+"""
 fillopacity(values::AbstractArray) =
         FillOpacity([FillOpacityPrimitive(value) for value in values])
 
--- a/src/svg.jl
+++ b/src/svg.jl
@@ -636,7 +636,7 @@ function print_property(img::SVG, property::LineWidthPrimitive)
 end
 
 function print_property(img::SVG, property::FillOpacityPrimitive)
-    print(img.out, " opacity=\"")
+    print(img.out, " fill-opacity=\"")
     svg_print_float(img.out, property.value)
     print(img.out, '"')
 end
@@ -711,9 +711,9 @@ function print_vector_properties(img::SVG, idx::Int, supress_fill::Bool=false)
 
         # let the opacity primitives clobber the alpha value in fill and stroke
         if propertytype == Fill && has_fill_opacity
-           print_property(img, FillPrimitive(RGBA{Float64}(property.primitives[idx].color, 1.0)))
+           print_property(img, FillPrimitive(RGBA{Float64}(color(property.primitives[idx].color), 1.0)))
         elseif propertytype == Stroke && has_stroke_opacity
-           print_property(img, StrokePrimitive(RGBA{Float64}(property.primitives[idx].color, 1.0)))
+           print_property(img, StrokePrimitive(RGBA{Float64}(color(property.primitives[idx].color), 1.0)))
         else
             print_property(img, property.primitives[idx])
         end
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -10,6 +10,7 @@ makedocs(
         "Tutorial" => "tutorial.md",
         "Gallery" => Any[
             "Forms" => "gallery/forms.md",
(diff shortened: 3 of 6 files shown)


src/property.jl
```diff
@@ -196,7 +208,18 @@ end
...
+"""
 fillopacity(value::Float64) = FillOpacity([FillOpacityPrimitive(value)])
+
+"""
+    fillopacity(values::AbstractArray)
+
+Arguments can be passed in arrays in order to perform multiple drawing operations at once.
+"""
 fillopacity(values::AbstractArray) =
         FillOpacity([FillOpacityPrimitive(value) for value in values])
```
bjarthur: is there a way that the dot broadcasting syntax could be used by the user instead of having to define a second method here?

bjarthur:
great to be fixing these old issues!  and to keep expanding the gallery.

my only other suggestion would be to add a minimal test.  ideally something that doesn't just reproduce the code in the gallery.

src/property.jl
Mattriks: No, because this is false:
```
typeof(fillopacity([0.5, 0.3])) .== typeof(fillopacity.([0.5, 0.3])) 
```

Mattriks:
Yes, I can add a minimal test.

Mattriks:
I tested this PR with Gadfly, and there are some side-effects:
```julia
Da = DataFrame(x=1:10, y=rand(10), c= rand(["a","b","c"],10))
p = plot(Da, x=:x, y=:y, color=:c,
    Geom.point, Geom.hair,
    Guide.title("Standard Plot"),
    Theme(major_label_color="red")
)
draw(PNG(6inch,3inch),p)
```
iss314_gfa
The `panel_fill` is now black, which is the default in `Compose`. This can be changed by making the default `panel_fill="white"` in Gadfly.
The reason for the missing y-axis and missing Title are less obvious, the next step is to investigate.

Mattriks:
I fixed all the above side-effects with the latest commit, and this next example provides further explanation about the black panel fill.
This example reproduces Gadfly's `Guide.background` [code](https://github.com/GiovineItalia/Gadfly.jl/blob/ddcf1a9fd7c5b9ab4fe3094ca6637d6889c5e014/src/guide.jl#L146-L152) (in panelb). In panelb, a black rectangle should appear, bacause `fill(nothing)` is the same as `fill(RGBA(0,0,0,0)`, and `fillopacity(1.0)` should change the alpha channel. But it doesn't on master:   

```julia
set_default_graphic_size(14cm,4cm)
panela = compose(context(order=1), rectangle(), fill(nothing), fillopacity(0.0) )
panelb = compose(context(order=1), rectangle(), fill(nothing), fillopacity(1.0) ) # Gadfly's current code
img  = compose(context(), hstack(panela, panelb))
```
compose_master
And with this PR, the same example gives, as expected:

compose_314





So the solution here is to make Gadfly's default `panel_opacity=0.0`, and then Gadfly plots will be the same as now. I will open a PR in Gadfly before finishing this PR - I still need to implement a miminal test here.

Mattriks:
I added a minimal test, rebased and squashed commits. Ready for merging.

tlnagy:
Awesome work @Mattriks!

src/svg.jl
```diff
@@ -711,9 +711,9 @@ function print_vector_properties(img::SVG, idx::Int, supress_fill::Bool=false)
 
         # let the opacity primitives clobber the alpha value in fill and stroke
         if propertytype == Fill && has_fill_opacity
-           print_property(img, FillPrimitive(RGBA{Float64}(property.primitives[idx].color, 1.0)))
+           print_property(img, FillPrimitive(RGBA{Float64}(color(property.primitives[idx].color), 1.0)))
```
tlnagy: This was a much better solution than mine in #315

tlnagy:
I'm running into a bug on Compose master after this PR merged. @Mattriks, what are your thoughts? Here's the data: [data.csv.zip](https://github.com/GiovineItalia/Compose.jl/files/2475685/data.csv.zip)

```julia
using CSV
using Gadfly

df2 = CSV.File("data.csv") |> DataFrame
plot(df2, x=:reltime, y=:volume, color=:sample,
     Geom.line, Geom.point,
     Guide.xlabel("Time (mins)"),
     Guide.ylabel("Volume (μm)^3"),
     Coulter.theme
)
```

I'm getting the following rendering error:

image

On line 78, I do see that fill-opacity is defined multiple times: 

```
      <g pointer-events="visible" fill-opacity="1" fill="#000000" fill-opacity="0.000" stroke="#000000" class="guide background" id="img-51165cf1-16">
```

tlnagy:
It looks like this might be happening because `fill-opacity` is printed both here:

https://github.com/GiovineItalia/Compose.jl/blob/567962d7a982103ffa45234d228d5d34473cb574/src/svg.jl#L595-L596

and here (added in this PR)

https://github.com/GiovineItalia/Compose.jl/blob/567962d7a982103ffa45234d228d5d34473cb574/src/svg.jl#L639

bjarthur:
i can confirm. https://github.com/GiovineItalia/Gadfly.jl/blob/master/test/testscripts/issue996.jl is another test case that fails.  we should wait to tag gadfly until this is fixed.

PR #316, opened 2018-10-13
fix for duplicate fill-opacity tags from #314

tlnagy:
This should fix the issue identified after the merger of #314 with duplicate fill-opacity tags.


tlnagy:
Reverted by #317

bjarthur:
please add a test for the second try!  :)

PR #319, opened 2018-10-16
Upgrade svg_fmt_color to svg rgba spec

### LINKED ISSUE #318: Fillopacity: new issues
@tlnagy, @bjarthur 
Here is a MWE of the subsequent #314 issue in Compose (this example reproduces the code in Gadfly's [`Guide.background`](https://github.com/GiovineItalia/Gadfly.jl/blob/3483498410a44a11512f1587bc9d91333d887b69/src/guide.jl#L146-L152)). Note the comments:
```julia
set_default_graphic_size(14cm,4cm)
panela = compose(context(order=1), rectangle(), fill(nothing), fillopacity(0.0) )
draw(SVG(), panela) # works in IJulia
draw(SVG("panela.svg"), panela) # opening file in browser shows error
```
The reason that error occurs is that `fill(nothing)` is RGBA(0,0,0,0), and hence in svg "fill-opacity=0" gets printed twice: once from `fill(nothing)` and once from `fillopacity(0.0)`.  
Perhaps a warning should be thrown in `Compose` - suggesting to the user that they can't use any xxxA fill color e.g. `RGBA(0,0,0,0)`and `fillopacity()` together for svg i.e. you can't specify the fill opacity twice for svg.
In Gadfly, there are two options to solve the issue:
i) make `theme.panel_fill` a e.g. RGBA color (it is), and deprecate `theme.panel_opacity` 
ii) make `theme.panel_fill` a e.g. RGB color, and use `theme.panel_opacity` to control opacity

Note that option i) is actually what is done in Gadfly's [`render(plot::Plot)` function](https://github.com/GiovineItalia/Gadfly.jl/blob/3483498410a44a11512f1587bc9d91333d887b69/src/Gadfly.jl#L756-L760), which has a `fill(nothing)` statement, and no `fillopacity()` statement, because `fill(nothing)` is `RGBA(0,0,0,0)`. 
Thoughts?

Mattriks:
Fixes #318
- [x] I've updated the documentation to reflect these changes
- [x] I've added and/or updated the unit tests
- [x] I've run the regression tests
- [x]  I've built the Compose docs and confirmed these changes don't cause new errors
- [x] I've run the Gadfly tests with this PR

### This PR 
- upgrades `svg_fmt_color()` to include a method for the svg `rgba` spec.
- This means opacity can be specified twice: in `fill(c)` and `fillopacity(a)`, producing the same result as in the svg backend (alphas multiply).

### DIFF
--- a/src/property.jl
+++ b/src/property.jl
@@ -211,7 +211,7 @@ const FillOpacity = Property{FillOpacityPrimitive}
 """
     fillopacity(value)
 
-Define a fill opacity, where 0≤value≤1.
+Define a fill opacity, where 0≤value≤1. Note for svg alphas multiply: e.g. using `fill(RGBA(r,g,b,a₁))` and `fillopacity(a₂)` together will result in an opacity=a₁*a₂. 
 """
 fillopacity(value::Float64) = FillOpacity([FillOpacityPrimitive(value)])
 
--- a/src/svg.jl
+++ b/src/svg.jl
@@ -85,6 +85,9 @@ end
 # Format a color for SVG.
 svg_fmt_color(c::Color) = string("#", hex(c))
 svg_fmt_color(c::Nothing) = "none"
+svg_fmt_color(c::CT) where CT<:TransparentColor =
+    string("rgba(", join(floor.(Int,[c.r,c.g,c.b].*255),","), ",", svg_fmt_float(c.alpha),")")
+
 
 # Replace newlines in a string with the appropriate SVG tspan tags.
 function svg_newlines(input::AbstractString, x::Float64)
@@ -591,12 +594,7 @@ function print_property(img::SVG, property::StrokePrimitive)
 end
 
 function print_property(img::SVG, property::FillPrimitive)
-    if property.color.alpha != 1.0
-        @printf(img.out, " fill=\"%s\" fill-opacity=\"%0.3f\"",
-                svg_fmt_color(color(property.color)), property.color.alpha)
-    else
-        @printf(img.out, " fill=\"%s\"", svg_fmt_color(color(property.color)))
-    end
+        @printf(img.out, " fill=\"%s\"", svg_fmt_color(property.color))
 end
 
 function print_property(img::SVG, property::StrokeDashPrimitive)
@@ -698,7 +696,6 @@ function print_vector_properties(img::SVG, idx::Int, supress_fill::Bool=false)
     end
 
     has_stroke_opacity = haskey(img.vector_properties, StrokeOpacity)
-    has_fill_opacity = haskey(img.vector_properties, FillOpacity)
 
     for (propertytype, property) in img.vector_properties
         if property === nothing ||
@@ -709,10 +706,8 @@ function print_vector_properties(img::SVG, idx::Int, supress_fill::Bool=false)
         idx > length(property.primitives) &&
                 error("Vector form and vector property differ in length. Can't distribute.")
 
-        # let the opacity primitives clobber the alpha value in fill and stroke
-        if propertytype == Fill && has_fill_opacity
-           print_property(img, FillPrimitive(RGBA{Float64}(color(property.primitives[idx].color), 1.0)))
-        elseif propertytype == Stroke && has_stroke_opacity
+        # let the opacity primitive clobber the alpha value in stroke
+        if propertytype == Stroke && has_stroke_opacity
            print_property(img, StrokePrimitive(RGBA{Float64}(color(property.primitives[idx].color), 1.0)))
         else
             print_property(img, property.primitives[idx])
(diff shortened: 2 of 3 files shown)


Mattriks:
In Gadfly, there are 2 tests that fail: line 15 & 16 in `testscripts/density_dark.jl`.  Note that all the generated images for `density_dark.jl` are fine.  The issue is that these tests test for hex strings in the svg string.  But in Compose the `FillPrimitive` automatically converts any color to a RGBA color, e.g. type  `fill("blue")`. This means that in Compose `svg.jl` any `fill(color)` will be converted to the rgba spec, rather than to a hex string. Hence in Gadfly the `density_dark.jl test` needs to be updated, after this PR has been merged (it's good to go!).

Note also that there is a second issue with the tests in `density_dark.jl`. The actual test suite gets repeated 6 times, which doesn't seem [what was intended](https://github.com/GiovineItalia/Gadfly.jl/blob/3483498410a44a11512f1587bc9d91333d887b69/test/testscripts/density_dark.jl#L11). I think this occurs because of the changes in scope.

src/property.jl
```diff
@@ -211,7 +211,7 @@ const FillOpacity = Property{FillOpacityPrimitive}
 """
     fillopacity(value)
 
-Define a fill opacity, where 0≤value≤1.
+Define a fill opacity, where 0≤value≤1. Note for svg alphas multiply: e.g. using `fill(RGBA(r,g,b,a₁))` and `fillopacity(a₂)` together will result in an opacity=a₁*a₂.
```
tlnagy:
```suggestion
Define a fill opacity, where 0≤value≤1. 

!!! info

    e.g. For SVG outputs, alpha values are multiplied together. For example, using `fill(RGBA(r,g,b,a₁))` and `fillopacity(a₂)` together will result in an opacity=a₁*a₂. 
```

tlnagy:
> In Gadfly, there are 2 tests that fail: [line 15 & 16](https://github.com/GiovineItalia/Gadfly.jl/blob/3483498410a44a11512f1587bc9d91333d887b69/test/testscripts/density_dark.jl#L15-L16) in `testscripts/density_dark.jl`. Note that all the generated images for `density_dark.jl` are fine. The issue is that these tests test for hex strings in the svg string. But in Compose the `FillPrimitive` automatically converts any color to a RGBA color, e.g. type `fill("blue")`. This means that in Compose `svg.jl` any `fill(color)` will be converted to the rgba spec, rather than to a hex string. Hence in Gadfly the `density_dark.jl test` needs to be updated, after this PR has been merged (it's good to go!).

I would be okay with updating the tests to not check hex codes directly

Mattriks:
Would anyone object to deleting [these lines](https://github.com/GiovineItalia/Gadfly.jl/blob/3483498410a44a11512f1587bc9d91333d887b69/test/testscripts/density_dark.jl#L11-L26)?  The images that the test generates still act as a test.

tlnagy:
What wrong with duplicating this line 
https://github.com/GiovineItalia/Compose.jl/blob/71ffc8e1399b30deb0546ade1cba681c4cfc7eec/src/svg.jl#L89
over in Gadfly to use instead of the hex codes? Those lines are really testing whether using the dark theme works, which is something we might miss with just the image generation (except when we run full regression testing). We really need to refactor how we do testing in Gadfly. @bjarthur, thoughts?

Mattriks:
Nothing wrong with upgrading those lines, just checking if people thought differently from me 🙂

bjarthur:
i'm fine with either deleting those lines or updating them to check for rgba.  i've tried to make regression testing painless enough to do with each PR, which i believe it is, just takes time.  in what way do you think gadfly testing needs refactoring?

bjarthur:
just delete those lines.  it's the only testscript which has `@test` inside, right?  so deleting them would make things more uniform across testscripts, and we'd just have to rely on the regression testing to make sure it worked right, which is fine.

bjarthur:
you could instead move them to the bottom of runtests.jl, or some other file which runtest.jl includes.  this is probably the best option, as we might want to add other tests like this as well.

bjarthur:
actually couldn't the loop in runtest.jl be re-written as:

```
@testset "Gadfly" begin
    for filename in testfiles
        Random.seed!(1)
        p = evalfile(joinpath(testdir, "$(filename).jl"))
        @test typeof(p) in [Plot,Compose.Context]
        for (backend_name, backend) in backends
            @info string(filename,'.',backend_name)
            r = draw(backend(filename), p)
            @test typeof(r) in [Bool,Nothing]
        end
    end
end
```

this would run the testscript only once, which would not only run the `@test` block in density_dark once, but tests overall should be faster this way as each `Plot` is only created once, and not re-created for each backend.

bjarthur:
@Mattriks @tlnagy  this PR has caused some [problems](https://discourse.julialang.org/t/gadfly-svgs-render-without-color/16804).

tlnagy:
It's not just a MS issue. I can reproduce with the latest Inkscape build on Ubuntu 18.04. We're gonna have to fix this regression. It's unfortunate that the RGBA standard isn't more widely supported.

The best option might be option ii from here https://github.com/GiovineItalia/Compose.jl/issues/318#issuecomment-429692039

Issue #322, opened 2018-10-27
Switch back from using RGBA for fill color 


tlnagy:
### Problem

As pointed out by https://discourse.julialang.org/t/gadfly-svgs-render-without-color/16804/, using RGBA with fill doesn't appear to be widely supported outside of browsers since it's not part of the SVG 1.1 spec. 

This is also noted in the [MDN SVG docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Fills_and_Strokes):

![image](https://user-images.githubusercontent.com/1661487/47609277-62564700-d9f0-11e8-9044-3493f459b4d0.png)

@bjarthur correctly noted that RGBA is permitted in the SVG2 spec (https://github.com/GiovineItalia/Compose.jl/issues/318#issuecomment-429625909), but that spec isn't widely adopted by major packages like Inkscape, MS Office, etc. I'm labeling this a bug since it's breaking and a high priority to fix.

### Solutions

The problem is that just reverting #319 isn't sufficient as then #318 will resurface. We might need to revert #314 as well until we get this worked out. @Mattriks thoughts? 

We need a good solution for https://github.com/GiovineItalia/Compose.jl/pull/314#issuecomment-429561491. The problem is that the vector `fill-opacity` is being printed without knowing whether a scalar `fill-opacity` is present so we end up with the double `fill-opacity` issue. Can the vector `fill-opacity` detect the presence of the other one and clobber it or multiply the opacities? The latter is preferable.

Mattriks:
I think the quick solution is to revert #319, and implement my suggestions prior to bjarthur's rgba suggestion: see my initial https://github.com/GiovineItalia/Compose.jl/issues/318#issue-369859071. My suggestions were for 2 changes: 
- In Compose: throw a warning if a user tries to use e.g. an RGBA color and `fillopacity` together.
- In Gadfly: probably option i: deprecate theme.panel_opacity

Mattriks:
Should I start by submitting a PR to Gadfly to deprecate `theme.panel_opacity`?

tlnagy:
> In Compose: throw a warning if a user tries to use e.g. an RGBA color and fillopacity together.

Ideally, we should be able to handle situations like this via multiplication of the two opacity values together, but I think this is an acceptable compromise in the mean time.

> Should I start by submitting a PR to Gadfly to deprecate `theme.panel_opacity`?

[...]

PR #324, opened 2018-10-28
revert to using hex for fill colors instead of RGBA

tlnagy:
2nd part of #322

### DIFF
--- a/src/property.jl
+++ b/src/property.jl
@@ -211,7 +211,7 @@ const FillOpacity = Property{FillOpacityPrimitive}
 """
     fillopacity(value)
 
-Define a fill opacity, where 0≤value≤1. Note for svg alphas multiply: e.g. using `fill(RGBA(r,g,b,a₁))` and `fillopacity(a₂)` together will result in an opacity=a₁*a₂. 
+Define a fill opacity, where 0≤value≤1.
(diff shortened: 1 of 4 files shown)


bjarthur:
bump

bjarthur:
is this ready to merge?  or is it stale?

Mattriks:
The function `svgalphatest` needs to be added to this PR.  See [this post](https://github.com/GiovineItalia/Compose.jl/issues/322#issuecomment-433741922) and the subsequent posts. If you merge this PR, I can add `svgalphatest` in another PR?

tlnagy:
As @Mattriks says, this PR needs to throw a warning to alert users to the change in behavior.

> If you merge this PR, I can add svgalphatest in another PR?

This is fine, IMO

bjarthur:
@Mattriks add the test and then we can tag

Mattriks:
The above merge had a side-effect: the `fill(nothing)` statement in a parent context now clobbers the fill(color) statement in a nested context (for svg).  See the 2 examples in the [Transformations gallery](http://giovineitalia.github.io/Compose.jl/dev/gallery/transforms/), which both use nested contexts.  I should be able to sort that out.

PR #326, opened 2018-11-06
Fix pango

vchuravy:
this adds minimal PNG support back, I very much wrote this in anger and havne't tested more than the few plots I desperatly needed.


vchuravy:
#325 was not enough for me.

vchuravy:
Ok this also now uses `Project.toml` files and an updated Travis config to get rid of some pesky "Cairo" not in your Project.toml.

PR #327, opened 2018-11-09
Requires.jl uses relative import

vchuravy:
Before:
```
julia> using Compose
julia> using Cairo
┌ Warning: Package Compose does not have Cairo in its dependencies:
│ - If you have Compose checked out for development and have
│   added Cairo as a dependency but haven't updated your primary
│ - Otherwise you may need to report an issue with Compose
└ Loading Cairo into Compose from project dependency, future warnings for Compose are suppressed.
```

After:
```
julia> using Compose

julia> using Cairo

julia> using Fontconfig
[ Info: Loading Fontconfig backend into Compose.jl
```


bjarthur:
thanks!

PR #340, opened 2019-01-17
added error messages for show methods involving Cairo

zsunberg:
I was initially dismayed to find that there is no `show(io, ::MIME"image/png", ::Context)`. It took me a while to figure out that I needed to import Cairo. This PR adds helpful error messages about that.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -122,9 +122,9 @@ default_line_width = 0.3mm
 default_stroke_color = nothing
 default_fill_color = colorant"black"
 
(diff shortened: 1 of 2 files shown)


bjarthur:
i'm fine with emitting an error on `show`, but there has to be a better way than to repeat in the code the error function almost verbatim.

zsunberg:
@bjarthur thanks for the feedback. I had a hard time getting rid of the repetition since the error messages are a little different. The latest commit has my best effort.

PR #341, opened 2019-01-27
Fix UndefVarError: pango_cairo_ctx not defined by Julia 1.x.

dmolina:
Dear authors, 

Using Gadfly, and Fontconfig I have found that error in Julia 1.x:

`
 Warning: [TEMPORARY WORKAROUND, pangolayout] for plotting with Gadfly.jl, see https://github.com/GiovineItalia/Gadfly.jl/issues/1206
└ @ Compose ~/.julia/packages/Compose/wlPCt/src/pango.jl:67
┌ Warning: Error requiring Fontconfig from Compose:
│ LoadError: UndefVarError: pango_cairo_ctx not defined
│ Stacktrace:
│  [1] Compose.PangoLayout() at /mnt/home/daniel/.julia/packages/Compose/wlPCt/src/pango.jl:59
│  [2] top-level scope at none:0
│  [3] include at ./boot.jl:326 [inlined]
│  [4] include_relative(::Module, ::String) at ./loading.jl:1038
│  [5] include at ./sysimg.jl:29 [inlined]
│  [6] include(::String) at /mnt/home/daniel/.julia/packages/Compose/wlPCt/src/Compose.jl:1
│  [7] top-level scope at /mnt/home/daniel/.julia/packages/Compose/wlPCt/src/Compose.jl:172
│  [8] eval at ./boot.jl:328 [inlined]
│  [9] eval at /mnt/home/daniel/.julia/packages/Compose/wlPCt/src/Compose.jl:1 [inlined]
│  [10] (::getfield(Compose, Symbol("##118#124")))() at /mnt/home/daniel/.julia/packages/Requires/9Jse8/src/require.jl:67
│  [11] err(::getfield(Compose, Symbol("##118#124")), ::Module, ::String) at /mnt/home/daniel/.julia/packages/Requires/9Jse8/src/require.jl:38
│  [12] #117 at /mnt/home/daniel/.julia/packages/Requires/9Jse8/src/require.jl:66 [inlined]
│  [13] withpath(::getfield(Compose, Symbol("##117#123")), ::String) at /mnt/home/daniel/.julia/packages/Requires/9Jse8/src/require.jl:28
│  [14] (::getfield(Compose, Symbol("##116#122")))() at /mnt/home/daniel/.julia/packages/Requires/9Jse8/src/require.jl:65
│  [15] #invokelatest#1 at ./essentials.jl:742 [inlined]
│  [16] invokelatest at ./essentials.jl:741 [inlined]
[...]

PR #342, opened 2019-01-31
Use Refs rather than globals for atsign-requires dependencies

timholy:
This appears to fix problems involving Compose, Cairo, and Fontconfig. The idea is to avoid using variables that get defined only via `global` statements that run only in `@require` blocks, and switch to a strategy to uses variables that are defined at precompilation time. Here I do that by defining the variables as `Ref`s, so I can update their contents when they get initialized.

Fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1206
Replaces https://github.com/GiovineItalia/Compose.jl/pull/341

This is not based on deep knowledge of what packages are or are not needed for my plotting task. I was simply following https://github.com/JuliaGraphs/GraphPlot.jl and then responded to messages about needing to `import Cairo, Fontconfig`.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -150,19 +150,20 @@ include("pgf_backend.jl")
 
 include("fontfallback.jl")
 
+const pango_cairo_ctx = Ref(C_NULL)
+const pango_cairo_fm = Ref(C_NULL)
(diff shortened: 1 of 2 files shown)


bjarthur:
thanks tim!

PR #345, opened 2019-03-03
Add documentation for PNG, PDF, PS

jerlich:
As discussed in GiovineItalia/Gadfly.jl#966

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -183,9 +183,30 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
             dpi = (B==PNGBackend ? 96 : 72)) where {B<:ImageBackend} =
         Image{B}(IOBuffer(), width, height, emit_on_finish, dpi=dpi)
 
+commondoc = raw"""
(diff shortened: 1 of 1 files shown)


src/cairo_backends.jl
```diff
@@ -183,10 +183,31 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
...
+```jldoctest
+    using Gadfly, Cairo
+    p = plot(x = 1:10, y=rand(10), Geom.line)
+    draw($(f)("myplot.$(lowercase(f))",10cm, 5cm, dpi=250),p)
+```
+"""
+
+for backend in [:PNG, :PDF, :PS]
+    str = docfunc(String(backend))
+    @eval @doc $str $backend
```
bjarthur: great to do this in a way which doesn't explicitly repeat the docstring three times!  how about going a step further and bringing the method definitions into the for loop too?  like [this](https://github.com/JaneliaSciComp/NIDAQ.jl/blob/master/src/analog.jl#L93).

jerlich:
This PR is incorporated in #347

Issue #175, opened 2015-12-11
Latex in the text form


jpfairbanks:
I see that if I draw on a PGF object, then I get a latex document that can be compiled into a PDF with latex text.

For example:
```julia
s = compose(context(),
(context(), text(1/2,1/2,"\$\\alpha=0.5\$")),
(context(), rectangle(), fill("bisque"))
)
draw(PGF("plot.tex", 6inch, 4inch), s)
```
![image](https://cloud.githubusercontent.com/assets/2753998/11732156/8e9b3bb8-9f6f-11e5-9d4a-8d191515a7de.png)
Although I have to delete the `\usepackage{fontspec}` line because I use pdflatex and not xelatex.

However in the IJulia notebook the latex expressions do not appear:
![image](https://cloud.githubusercontent.com/assets/2753998/11732197/de05c61e-9f6f-11e5-9455-3fc559aa4bac.png)
Plain text works fine in the IJulia notebook.

Is there a way to get the latex expressions rendered properly in the IJulia notebook?

darwindarak:
Not at the moment, unfortunately.  The output to the notebooks are in SVG format by default, and we do have have a Julia implementation of LaTeX's math layout engine *yet*.

As for disabling the `fontspec` option, you should be able to use the `texfonts=true` option in the `PGF` call (e.g. `PGF("plot.tex", 6inch, 4inch, texfonts=true`).

PR #347, opened 2019-03-04
docs for PNG, PS, PDF, SVG, PGF

jerlich:
+ Took suggestion from previous PR to put methods declarations in the loop
+ Added custom doc string for SVG and PGF, since they have different options.
+ note: PGF example uses `.tex` extension. Is that correct?

### DIFF
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -186,10 +186,10 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
 
 
 docfunc(f) = """
-    $f([out::Union{AbstractString, IOBuffer}, width=√200cm, height=10cm; dpi=72])
+    $f([out::Union{AbstractString, IOBuffer}], width=√200cm, height=10cm; dpi=$(f==:PNG ? 96 : 72))
 
 Create a [`Compose.$(f)backend`](@ref) with an associated file and specified width, height (as positional arguments)
-and dpi (as a keyword argument). For PNG, the default dpi is 96, otherwise 72. The first argument is optional and can be an [`IOBuffer`](@ref). 
+and dpi (as a keyword argument). The first argument is optional and can be an [`IOBuffer`](@ref). 
 Normally passed to [`draw`](@ref). 
 
 # Examples
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -183,9 +183,9 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
             dpi = (B==PNGBackend ? 96 : 72)) where {B<:ImageBackend} =
         Image{B}(IOBuffer(), width, height, emit_on_finish, dpi=dpi)
 
-PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
-PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
-PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
+#PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
+#PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
+#PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
 
 
 
@@ -199,15 +199,20 @@ and dpi (as a keyword argument). Normally passed to [`draw`](@ref).
 ```jldoctest
     using Gadfly, Cairo
     p = plot(x = 1:10, y=rand(10), Geom.line)
-    draw($(f)("myplot.$(lowercase(f))",10cm, 5cm, dpi=250),p)
+    draw($(f)("myplot.$(lowercase(String(f)))",10cm, 5cm, dpi=250),p)
 ```
 """
 
 
-for backend in [:PNG, :PDF, :PS]
-    str = docfunc(String(backend))
-    @eval @doc $str $backend
+for func in [:PNG, :PDF, :PS]
+    backend = Symbol(func, "Backend")
+    docstr = docfunc(func)
+    @eval $func(args...; kwargs...) = Image{$backend}(args...; kwargs...)
+    @eval @doc $docstr $func
+
 end
+
+
 
 const CAIROSURFACE = Image{CairoBackend}
 
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -193,7 +193,7 @@ docfunc(f) = """
     $f(filename::AbstractString, width::MeasureOrNumber, height::MeasureOrNumber; dpi)
 
 Create a [`Compose.$(f)backend`](@ref) with an associated file and specified width, height (as positional arguments)
-and dpi (as a keyword argument). Normally passed to [`draw`](@ref).
+and dpi (as a keyword argument). For PNG, the default dpi is 96, otherwise 72. Normally passed to [`draw`](@ref). 
 
 # Examples
 ```jldoctest
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -187,6 +187,28 @@ PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
 PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
 PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
 
+
+
+docfunc(f) = """
+    $f(filename::AbstractString, width::MeasureOrNumber, height::MeasureOrNumber; dpi)
+
+Create a [`Compose.$(f)backend`](@ref) with an associated file and specified width, height (as positional arguments)
+and dpi (as a keyword argument). Normally passed to [`draw`](@ref).
+
+# Examples
+```jldoctest
+    using Gadfly, Cairo
+    p = plot(x = 1:10, y=rand(10), Geom.line)
+    draw($(f)("myplot.$(lowercase(f))",10cm, 5cm, dpi=250),p)
+```
+"""
+
+
+for backend in [:PNG, :PDF, :PS]
+    str = docfunc(String(backend))
+    @eval @doc $str $backend
+end
+
 const CAIROSURFACE = Image{CairoBackend}
 
 function (img::Image)(x)
--- a/src/cairo_backends.jl
+++ b/src/cairo_backends.jl
@@ -183,17 +183,14 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
             dpi = (B==PNGBackend ? 96 : 72)) where {B<:ImageBackend} =
         Image{B}(IOBuffer(), width, height, emit_on_finish, dpi=dpi)
 
-#PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
-#PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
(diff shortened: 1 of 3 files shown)


Mattriks:
Point 3 above: yes, see e.g. #175.
Not sure why you opened a new PR here, rather than making changes to #345.

jerlich:
re: editing PR. sorry, didn't know that was preferred.

Mattriks:
In the Julia 0.6 docs:
https://docs.julialang.org/en/v0.6/manual/packages/#Making-changes-to-an-existing-package-1
There is some useful info there (e,g, squashing and rebasing).
I couldn't find the same section in Julia 1.0 docs.

jerlich:
Are you waiting for me to move these into the #345? Should I close this?

Mattriks:
Since this PR is more advanced, I think just close #345.

src/pgf_backend.jl
```diff
@@ -119,6 +119,27 @@ function PGF(out::IO,
 end
 
 # Write to a file.
+"""
+    PGF(filename::AbstractString, width::MeasureOrNumber, height::MeasureOrNumber, only_tikz::Bool; texfonts::Bool)
+
+Create a [`Compose.PGF`](@ref) with an associated file and specified width, height and only_tikz 
+(as positional arguments) and texfonts (as a keyword argument)
+Normally passed to [`draw`](@ref). First argument can be an [`IOBuffer`](@ref)
```
bjarthur: the cairo backends (PDF, etc.) can also input an [IO buffer](https://github.com/GiovineItalia/Compose.jl/blob/master/src/cairo_backends.jl#L151).  should mention that there too.

and a better way to do this might be in the function signature instead of the text description.  something like

```
  PGF([output::Union{IO, AbstractString}], width...)
```

where the square brackets indicate it's optional, and the Union means you can use either.

src/pgf_backend.jl
```diff
@@ -119,6 +119,27 @@ function PGF(out::IO,
...
+
+# Arguments
++ `filename`: A filename or an IOBuffer
++ `width=√200cm`: PGF width (~14cm)
++ `height=10cm`: PGF height
++ `only_tikz=false`: Controls whether the latex output generates a complete latex document with headers/footer or only "tikzpicture" header/footers 
++ `texfonts=false`: If false include "\\usepackage{fontspec}" in the headers.
```
bjarthur: an explicit arguments section is generally avoided.  see point 4 [here](https://docs.julialang.org/en/latest/manual/documentation/).  maybe put this info in with the text you already have above?

src/cairo_backends.jl
```diff
@@ -183,9 +183,36 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
             dpi = (B==PNGBackend ? 96 : 72)) where {B<:ImageBackend} =
         Image{B}(IOBuffer(), width, height, emit_on_finish, dpi=dpi)
 
-PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
-PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
-PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
+#PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
+#PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
+#PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
```
bjarthur: no need to keep these around.  just delete these lines instead of comment them.

bjarthur:
looking great!  thanks for incorporating the previous suggestions.  i've made a few more.  if you have the time to include them too, the way to modify this PR is to:

```
git add
git commit
git rebase -i
git push --force github <branchname>
```

src/pgf_backend.jl
jerlich: I know that explicit argument lists are not preferred but they are not forbidden. It seems to me easier to explain the defaults and also the not obvious meaning of only-tikz and texfonts.

src/cairo_backends.jl
jerlich: yes. I intended to remove them. My bad. Will do.

src/pgf_backend.jl
jerlich: Is the first argument optional? I thought it has to be either a file name or IObuffer. What’s the example use case for calling it without any arguments?

bjarthur: yes it's optional.  see [this method](https://github.com/GiovineItalia/Compose.jl/blob/master/src/svg.jl#L275).  to be honest, i'm not sure what the used case is, but the method exists and adding square brackets to the docs is easy.

src/pgf_backend.jl
bjarthur: the defaults can be put in the function signature with equals signs [like this](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/geom/beeswarm.jl#L7).  note too how in the [corresponding docstring](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/geom/beeswarm.jl#L16) for this type the arguments are explained with narrative instead of bullet points.

i find arguments sections to use up too much screen real estate.  don't want to bike shed you to death though so do what you think is best and i'll merge.

jerlich: I agree the arguments take up a lot of space in the REPL. I get that the Julia idiom is "narrative" docs...not sure I like that decision, but I guess I have to learn to accept the things I cannot change ;).  I have a deadline looming, but I'll make it more idiomatic in the next week or so.

src/pgf_backend.jl
Mattriks: One use case is for jupyter notebooks. You can find examples in past Gadfly issues e.g. GiovineItalia/Gadfly.jl#936

jerlich:
sorry for the long delay. Is this PR still "Live"?

bjarthur:
yes!

jerlich:
ok. I have made all the changes (I hope!) through the GitHub web UI. Does the PR automatically reflect the most up to date version of the branch?

src/cairo_backends.jl
```diff
@@ -183,9 +183,33 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
             dpi = (B==PNGBackend ? 96 : 72)) where {B<:ImageBackend} =
         Image{B}(IOBuffer(), width, height, emit_on_finish, dpi=dpi)
 
-PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
-PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
-PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
+
+
+docfunc(f) = """
+    $f([out::Union{AbstractString, IOBuffer}, width=√200cm, height=10cm; dpi=72])
```
bjarthur: the 72 here could be replaced with `f==:PNG ? 96 : 72`, and the corresponding line in the narrative below deleted (ie "For PNG, the default..."

bjarthur: also, the trailing ] should be after the first argument, as it's obvious that the others are optional since they either have defaults or are keywords.

same applies to the SVG docstring.

and the square brackets are missing entirely from PGF.

src/pgf_backend.jl
```diff
@@ -119,6 +119,22 @@ function PGF(out::IO,
...
+
+Create a [`Compose.PGF`](@ref) with an associated file and specified width, height and only_tikz 
+(as positional arguments) and texfonts (as a keyword argument)
+Normally passed to [`draw`](@ref). First argument can be an [`IOBuffer`](@ref). If `only_tikz=true` 
+then the output is a "tikzpicture" otherwise the output is a commplete latex document with headers/footer. 
+If `texfonts=false`, include "\\usepackage{fontspec}" in the headers, otherwise skip that include.
+
+# Examples
+```jldoctest
+    using Gadfly, Cairo
```
bjarthur: Cairo is not needed for PGF

...or SVG.

one of the huge advantages of Gadfly is that it is pure julia for those two.

jerlich:
Ok. I think these last few  commits fix everything.

jerlich:
The travis-ci check is failing on installing Cairo on Mac, it seems.

src/cairo_backends.jl
```diff
@@ -183,9 +183,33 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
...
+docfunc(f) = """
+    $f([out::Union{AbstractString, IOBuffer}], width=√200cm, height=10cm; dpi=$(f==:PNG ? 96 : 72))
+
+Create a [`Compose.$(f)backend`](@ref) with an associated file and specified width, height (as positional arguments)
+and dpi (as a keyword argument). The first argument is optional and can be an [`IOBuffer`](@ref). 
+Normally passed to [`draw`](@ref). 
+
+# Examples
+```jldoctest
+    using Gadfly, Cairo
```
bjarthur: doh!  just realized this is a PR to Compose, not Gadfly, so the examples here and elsewhere should be something along the lines of `c = compose(context(), ellipse()); draw(..., c)`, with no reference to Gadfly or `plot`.  there are several packages which use Compose directly, bypassing Gadfly.

src/cairo_backends.jl
```diff
@@ -183,9 +183,33 @@ Image{B}(width::MeasureOrNumber=default_graphic_width,
...
-PNG(args...; kwargs...) = Image{PNGBackend}(args...; kwargs...)
-PDF(args...; kwargs...) = Image{PDFBackend}(args...; kwargs...)
-PS(args...; kwargs...) = Image{PSBackend}(args...; kwargs...)
+
+
+docfunc(f) = """
+    $f([out::Union{AbstractString, IOBuffer}], width=√200cm, height=10cm; dpi=$(f==:PNG ? 96 : 72))
+
+Create a [`Compose.$(f)backend`](@ref) with an associated file and specified width, height (as positional arguments)
+and dpi (as a keyword argument). The first argument is optional and can be an [`IOBuffer`](@ref).
```
bjarthur: would be good to mention `out` can be a filename too, as you did for SVG.

src/pgf_backend.jl
```diff
@@ -119,6 +119,22 @@ function PGF(out::IO,
 end
 
 # Write to a file.
+"""
+    PGF(filename::AbstractString, width=√200cm, height=10cm, only_tikz=false; texfonts=false)
```
bjarthur: first arg is optional here too, so [] is warranted, and the type is a Union with IO just like SVG, PDF, etc.

src/pgf_backend.jl
```diff
@@ -119,6 +119,22 @@ function PGF(out::IO,
...
+
+Create a [`Compose.PGF`](@ref) with an associated file and specified width, height and only_tikz 
+(as positional arguments) and texfonts (as a keyword argument)
+Normally passed to [`draw`](@ref). First argument can be an [`IOBuffer`](@ref). If `only_tikz=true`
```
bjarthur: again, explicitly mentioning a filename would be great.

bjarthur:
i've checkout out this PR and addressed my comments above.  will merge soon.  thanks for you help with this!

bjarthur:
closed by https://github.com/GiovineItalia/Compose.jl/commit/f5aac34fb77f37c17fbfbbb77dd080442a26425a

src/cairo_backends.jl
jerlich: Is there a way to have docs for these functions in Gadfly? I think many users use these as part of Gadfly, so it makes sense to have examples.

PR #352, opened 2019-05-12
get docs to build again

bjarthur:
thanks to @mortenpi for the `SVGJSWriteable` trick!  necessary here because Cairo does not support bitmaps.


lobingera:
@bjarthur #141 would provide it

bjarthur:
@lobingera i forgot about that PR! would be great if you had the time to rebase.

PR #354, opened 2019-07-09
Remove vestigial method

cmcaine:


cmcaine:
As mentioned here: https://github.com/GiovineItalia/Gadfly.jl/pull/1295

Tests pass before and after.

PR #365, opened 2019-11-06
CI Update

CiaranOMara:
- ~~Add Julia 1.3 to test matrix.~~
- Use the codecov flag in Travis instead of the manual code version.
- Pkg.instantiate after  Pkg.develop.
- Add colour to make doc output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,24 @@
 language: julia
+
 os:
     - linux
     - osx
     - windows
-sudo: required
-dist: trusty
+
 julia:
     - 1.0
     - 1.2
     - nightly
+
 matrix:
+  fast_finish: true
   allow_failures:
   - julia: nightly
+
 notifications:
     email: false
-after_success:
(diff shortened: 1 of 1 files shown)


bjarthur:
i'd suggest waiting to merge this until after julia 1.3 comes out, and then not testing compose on 1.2 anymore.  it might be awhile, as there is an [RC5 planned](https://github.com/JuliaLang/julia/pull/33630).

CiaranOMara:
I removed Julia v1.3 from the test matrix so as not to be held up by RC5.

bjarthur:
was doing `instantiate` before `develop` causing a problem?  i can't think of a reason why the order would matter.

CiaranOMara:
I don't think the order would have caused any issues as both `Pkg.develop()` and `Pkg.instantiate()` should upgrade or downgrade packages as required, and I assume `Pkg.jl` would arrive at the same solution irrespective of order. In the end, this just matches the order in the [Documenter.jl hosting example](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.travis.yml-Configuration-1).

PR #366, opened 2019-11-06
Set Cairo.jl compatibility

CiaranOMara:
This pull request is in response to https://github.com/GiovineItalia/Gadfly.jl/pull/1339#issuecomment-550114393. 

I did not get a diff between the test output with Cairo.jl compatibility set to v0.7 and with Cairo.jl compatibility explicitly set to v0.8.


CiaranOMara:
The documentation built from this commit also looks good to me.

bjarthur:
thanks for this!

PR #379, opened 2020-01-04
Bezigon Primitive

Mattriks:
- [x] I've updated the documentation to reflect these changes
- [x] I've built the docs and confirmed these changes don't cause new errors
- [x] I've added and/or updated the unit tests
- [x] I've run the regression tests

### This PR
- adds a [bezigon](https://en.wikipedia.org/wiki/Composite_B%C3%A9zier_curve) form
- Suggested in https://github.com/GiovineItalia/Compose.jl/pull/298#issuecomment-531218947

## To do

- [x] Implement `bezigon` for the PGF backend.
- [x] Support quadratic beziers by transforming to cubic (for Cairo)

## Example

```julia
petal = [[(0.4, 0.4), (0.4, 0.2), (0.5, 0.0)],  [(0.6, 0.2), (0.6, 0.4), (0.5, 0.5)]]
petalf(θ::Float64) =    (context(rotation=Rotation(θ, 0.5,0.5)),
    bezigon((0.5, 0.5), petal), fill(LCHuvA(70.,50., 360*θ/2π, 0.4)) )

 theta = range(π/20, 2π, step=2π/10) .-π
 compose(context(), petalf.(theta)...)
```
petals

### DIFF
--- a/docs/src/gallery/forms.md
+++ b/docs/src/gallery/forms.md
@@ -23,6 +23,20 @@ img2 = compose(context(),
 hstack(img1, img2)
 ```
 
+## [`bezigon`](@ref)
+ 
+```@example
+using Colors, Compose
+set_default_graphic_size(14cm,10cm)
+
+petal = [[(0.4, 0.4), (0.4, 0.2), (0.5, 0.0)],  [(0.6, 0.2), (0.6, 0.4), (0.5, 0.5)]]
+petalf(θ::Float64) = (context(rotation=Rotation(θ, 0.5,0.5)),
+    bezigon((0.5, 0.5), petal), fill(LCHuvA(70.,50., 360*θ/2π, 0.4)))
+
+theta = range(π/20, 2π, step=2π/10).-π
+img = compose(context(), petalf.(theta)...)
+```
+
 ## [`bitmap`](@ref)
 ```@example
 using Main: SVGJSWritable #hide
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -18,7 +18,7 @@ import Measures: resolve, w, h
 export compose, compose!, Context, UnitBox, AbsoluteBoundingBox,
         Rotation, Mirror, Shear,
        ParentDrawContext, context, ctxpromise, table, set_units!, minwidth, minheight,
-       text_extents, max_text_extents, polygon, ngon, star, xgon,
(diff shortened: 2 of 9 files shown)


Mattriks:
@Vectornaut  please test this branch & provide feedback.

tlnagy:
What about using that as Compose.jl's logo? It looks really nice :)

Mattriks:
Now more complex shapes are possible:
```julia
spade = [[(11,20), (13,20)], [(7,20)], [(9,20), (9,15)],
    [(9,20),(0,21),(0,16)], [(0,11), (6,9), (10,0)], 
    [(14,9), (20,11), (20,16)], [(20,21), (11,20), (11,15)]
]

compose(context(units=UnitBox(0,0,20,20)), 
        (context(), bezigon((11,15), spade), fill("gray"))
)
```
spade

Mattriks:
Done. I think the Compose logo could be changed in a separate PR.

Vectornaut:
@Mattriks, thank you so much for your hard work on this feature! I'm sorry I've taken so long to post feedback. The `bezigon` primitive works perfectly in the code I [commented about](https://github.com/GiovineItalia/Compose.jl/pull/298#issuecomment-531116925) in #298, which used the `path` primitive before. Its output is visually indistinguishable from what I had before.

I did some performance comparisons using two benchmark examples. Significant differences are in bold. Time and memory use are measured in repeated runs, ignoring the initial compilation run.

When I just replaced `path` with `bezigon`, not taking advantage of any other new features, here's what happened.

- Run time **fell by 60%.**
- Memory use **fell by 40%.**
- For the first example, the PDF output file grew by 8%. For the second example, it stayed the same size.
- For the first example, the SVG output file **grew by 50%.**

A big advantage of `bezigon` over `path` is that it can be vectorized! When I reorganized my code so I could use vectorized `bezigon` instead of ordinary `bezigon`, here's what happened.

- Run time **fell by 85%**.
- Memory use **fell by 85%**.
- For both examples, the PDF output shrank by 1%.
- For the first example, the SVG output stayed the same size.

The huge runtime performance gains are great for me. Now I can draw finer details and longer animations! The substantially larger SVG output file is weird, but no problem for me so far. In particular, it doesn't seem to slow down Inkscape's handling of the SVG.

PR #400, opened 2020-06-24
Compose v0.9

Mattriks:


bjarthur:
there are breaking changed that merit a bump in the minor number?

Mattriks:
This is needed for GiovineItalia/Gadfly.jl#1459, and I thought minor releases could include new features, such as the handling of missing values (which is an important step for Compose and Gadfly).  I can change to a patch release, if you think that is sufficient?

Mattriks:
Also, a user could write code for Compose 0.9, which is not compatible with Compose 0.8.x, because with Compose 0.9 I could remove `skipmissing` from code, but that would fail on Compose 0.8.x.

Mattriks:
@bjarthur I see your thumbs up - great! I will merge and register

bjarthur:
IIRC before 1.0 in semvar anything goes in terms of features and compatibility when bumping the major, minor, or patch numbers.  given that flexibility, i'd suggest the next version of Compose after the just-merged 0.9 doesn't necessary have to be 1.0.  better in my mind that it be 0.10.0.  having restrictions on backward compatibility is a pain.  it was a mistake in my mind to release Gadfly 1.0 so soon.  ditto for julia 1.0.

Mattriks:
I agree, minor versions of Compose should continue with 0.10.x 0.11.x etc.  There are some things in Compose which still need a bit of work, as can be seen on the issues page.

PR #405, opened 2020-10-06
Update typo in docs for `text` function in form.jl

senritsu:
fixed a typo I stumbled across while using Pluto


bjarthur:
thanks!

PR #407, opened 2020-12-23
Remove unused code

rikhuijzer:
Correct me if I'm wrong, but I cannot think of any way how these lines of code are actually doing anything :)

### DIFF
--- a/src/table-jump.jl
+++ b/src/table-jump.jl
@@ -143,12 +143,9 @@ function realize(tbl::Table, drawctx::ParentDrawContext)
             force_aspect_ratio!(tbl, x_solution, y_solution, w_solution, h_solution)
 
     # set child positions according to layout solution
-    feasible_eps = 1e-4
-    feasible = true
     for i in 1:m, j in 1:n
         if length(tbl.children[i, j]) == 1
             ctx = copy(tbl.children[i, j][1])
-            feasible == feasible && issatisfied(ctx, w_solution[j], h_solution[i])
             ctx.box = BoundingBox(
                 x_solution[j]*mm, y_solution[i]*mm,
                 w_solution[j]*mm, h_solution[i]*mm)
@@ -159,15 +156,12 @@ function realize(tbl::Table, drawctx::ParentDrawContext)
     for (l, (i, j, k)) in enumerate(c_indexes)
         if round(c_solution[l]) == 1
             ctx = copy(tbl.children[i, j][k])
-            feasible == feasible && issatisfied(ctx, w_solution[j], h_solution[i])
             ctx.box = BoundingBox(
                 x_solution[j]*mm, y_solution[i]*mm,
                 w_solution[j]*mm, h_solution[i]*mm)
             compose!(root, ctx)
         end
     end
-
-    feasible || warn("Graphic may not be drawn correctly at the given size.")
 
     return root
 end
--- a/src/table-jump.jl
+++ b/src/table-jump.jl
@@ -145,9 +145,12 @@ function realize(tbl::Table, drawctx::ParentDrawContext)
             force_aspect_ratio!(tbl, x_solution, y_solution, w_solution, h_solution)
 
     # set child positions according to layout solution
+    feasible_eps = 1e-4
+    feasible = true
     for i in 1:m, j in 1:n
         if length(tbl.children[i, j]) == 1
             ctx = copy(tbl.children[i, j][1])
+            feasible = feasible && issatisfied(ctx, w_solution[j], h_solution[i])
             ctx.box = BoundingBox(
                 x_solution[j]*mm, y_solution[i]*mm,
                 w_solution[j]*mm, h_solution[i]*mm)
@@ -166,7 +169,7 @@ function realize(tbl::Table, drawctx::ParentDrawContext)
         end
     end
 
-    feasible || warn("Graphics may not be drawn correctly at the given size.")
+    feasible || warn("Graphic may not be drawn correctly at the given size.")
 
     return root
 end
--- a/src/table-jump.jl
+++ b/src/table-jump.jl
@@ -106,6 +106,8 @@ function realize(tbl::Table, drawctx::ParentDrawContext)
     end
 
     # minimum cell size constraint for fixed cells
+    feasible_eps = 1e-4
+    feasible = true
     for i in 1:m, j in 1:n
         if length(tbl.children[i, j]) == 1
             minw = minwidth(tbl.children[i, j][1])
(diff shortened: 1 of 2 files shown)


bjarthur:
can `issatisfied()` really never return `false`?

rikhuijzer:
> can `issatisfied()` really never return `false`?

I don't think that the output of `issatisfied()` could alter the behavior of the program, because it isn't written to a variable nor is it returned in the function call. The only possibility would be if `issatisfied()` is not a pure function call because that means it would change state somewhere else. However, the definition of `issatisfied` is
```
# src/table.jl line 111 - 115
function issatisfied(ctx::Context, width, height)
    eps = 1e-4
    return (minwidth(ctx) == nothing || width + eps >= minwidth(ctx)) &&            
    (minheight(ctx) == nothing || height + eps >= minheight(ctx))
end
```
where `eps` is the only non-const occurrence of `eps` that I can find, so it must be a local variable, and minwidth and minheight are defined as
```
# src/container.jl line 128 - 129
minwidth(cont::Container) = cont.minwidth
minheight(cont::Container) = cont.minheight
```
which is a getter on `mutable struct Context <: Container`.

rikhuijzer:
> can `issatisfied()` really never return `false`?

I don't think that the output of `issatisfied()` could alter the behavior of the program, because it isn't written to a variable nor is it returned in the function call. The only possibility would be if `issatisfied()` is not a pure function call because that means it would change state somewhere else. However, the definition of `issatisfied` is
```
# src/table.jl line 111 - 115
function issatisfied(ctx::Context, width, height)
    eps = 1e-4
    return (minwidth(ctx) == nothing || width + eps >= minwidth(ctx)) &&            
    (minheight(ctx) == nothing || height + eps >= minheight(ctx))
end
```
where `eps` is the only non-const occurrence of `eps` that I can find, so it must be a local variable, and minwidth and minheight are defined as
```
# src/container.jl line 128 - 129
minwidth(cont::Container) = cont.minwidth
minheight(cont::Container) = cont.minheight
```
which is are both getters (and not setters) on  `mutable struct Context <: Container`.

rikhuijzer:
When applying this PR, all tests still pass on Julia 1. Julia 1.0 fails on the compatibility requirements.

@bjarthur Are the problems with Cairo also the reason why Gadfly and Compose development seems to be slowed down?

bjarthur:
@Mattriks @tlnagy thoughts on this PR?  i'm not familiar enough with this section of code to know if there are unforseen ramifcations.

@rikhuijzer did the Gadfly tests also pass?  the actual tests of Compose are quite limited, and we really rely mostly on Gadfly's tests to ensure Compose works.  in particular, the [regression tests](http://gadflyjl.org/dev/dev/regression/).

re. dev slowing down.  @Mattriks has been carrying the torch mostly lately.  i doubt cairo is an issue.  i actually mostly use the SVG backend which is pure julia and no cairo, in part because of the [interactivity](http://gadflyjl.org/dev/tutorial/#Interactivity-1).  i too prefer gadfly over the others (though i love UnicodePlots!).  most people like Plots though, though i wonder whether that is simply because the name implies that it's officially endorsed.  would be great to have more contributions from you!

Mattriks:
The `table-jump.jl` code is not used (see [these lines](https://github.com/GiovineItalia/Compose.jl/blob/d58b6902d425d722e53a14600ea66515c6dccea5/src/table.jl#L345-L349)) so changing it seems a moot point.  I'll have a look at the `table.jl` code sometime soon to see if those lines could be improved.  

Gadfly development has only slowed down because I've been busy with other work over the past 4 months (see [Gadfly commits](https://github.com/GiovineItalia/Gadfly.jl/commits/master)).  Hoping to get back to it in the new year!

Mattriks:
My assessment is that in `table.jl` there's a typo. [This  line](https://github.com/GiovineItalia/Compose.jl/blob/d58b6902d425d722e53a14600ea66515c6dccea5/src/table.jl#L333) should be: 
```julia
feasible = feasible && issatisfied(ctx, w_solution[j], h_solution[i])
```
That way, `feasible = true` until `issatisfied(....) = false`, then `feasible` remains false, triggering the warning after the loop (updated):
```julia
feasible || @warn "Graphic may not be drawn correctly at the given size."
```
In Gadfly, I was able to trigger this warning when I made a plot too small:
```julia
p = plot(x=1:5, y=rand(5), Geom.line)
draw(PNG(4cm, 4cm), p) # works
draw(PNG(3cm, 3cm), p) # warning
```

rikhuijzer:
@bjarthur For me, tests for Gadfly.jl and Cairo.jl fail locally because Cairo fails to precompile. This is probably because I'm running NixOS. I still have to find a workaround for this, and am planning to set-up a (Docker) dev container, which I'm not particularly fond of. 

> would be great to have more contributions from you!

Thank you. I must say that PRs like this are helping me in understanding Gadfly/Compose better. Thank you also for pointing me towards the documentation.

@Mattriks thanks for looking into it in detail! Now the code makes much more sense.

I've made the PR such that it includes the fix as offered by Mattriks. If merging, please do squash because the commits are a mess again :smile:

src/table-jump.jl
```diff
@@ -148,7 +148,7 @@ function realize(tbl::Table, drawctx::ParentDrawContext)
     for i in 1:m, j in 1:n
         if length(tbl.children[i, j]) == 1
             ctx = copy(tbl.children[i, j][1])
-            feasible == feasible && issatisfied(ctx, w_solution[j], h_solution[i])
```
Mattriks, line 151: Really just a comment, I don't see the point in changing code which is currently unused.  Re-including `table-jump.jl` in Compose (and updating the code) should be done in a separate PR. I'm also unconvinced that Jump optimization would consistently work better than the brute force approach currently used in `table.jl`.

src/table.jl
```diff
@@ -330,7 +330,7 @@ function realize_brute_force(tbl::Table, drawctx::ParentDrawContext)
             idx = optimal_choice[(j-1)*m + i]
             ctx = copy(tbl.children[i, j][idx])
         end
-        feasible == feasible && issatisfied(ctx, w_solution[j], h_solution[i])
```
Mattriks, line 333: You need to also update L340 below, because `warn` was deprecated in Julia 0.7 (JuliaLang/julia#24490), as in my previous post.

src/table-jump.jl
rikhuijzer: I've changed this part because it's clearly a clone of the code which contains a mistake as you have pointed out. As I see it, the benefit of changing unused code is to avoid people to bump their heads against this typo at a (much) later point in time.

src/table.jl
rikhuijzer: I didn't change it yet in table-jump because that change still depends on your response on the discussion about `table-jump.jl`. Also, fixing warn is not as important because it will either give a deprecation warning or fail to run on newer Julia versions.

Mattriks:
I ran the Gadfly tests with this PR and found 8 testscripts which generate the warning, mainly due to the size of the test plots being slightly undersized.

Here's one that looks ok to me, but still generates the warning:
```julia
p5 = evalfile(dirname(pathof(Gadfly))*"/../test/testscripts/scale_alpha_discrete.jl")
draw(PNG(), p5)
```
However the warning does seem to hint at some uncertainty! 

I also noticed in Jupyter that when I don't include the `draw` statement, the warning is printed in triple (I don't know the reason). 
I wonder if the warning needs to add some helpful info about resizing?

rikhuijzer:
I won't be able to work on this PR today. If you want, you should be able to edit my changes directly (I've set the checkmark in GitHub to allow edits by maintainers.) Alternatively, close this PR and create your own one :) Please don't let me being unavailable today hinder your work :smile:

lobingera:
@rikhuijzer I read this just by coincidence, could you please file an issue to Cairo.jl on precompilation.

PR #408, opened 2020-12-23
Switch to GitHub Actions

rikhuijzer:
Unlike the one at Gadfly, this one also runs on Windows and MacOS because the travis CI was also doing that. This means that 3 * 3 = 9 jobs will run, which is a bit much. Maybe we should remove `nightly`?


bjarthur:
i'm fine with removing nightly.

rikhuijzer:
@bjarthur done

bjarthur:
thanks!

PR #418, opened 2021-02-16
new mathjax support, requires JS

GiggleLiu:
# List of changes
## MathJax support when rendering SVG in a browser

It works directly in Pluto notebooks.
```julia
compose(context(), mathjax(0.2, 0.2, 0.1, 0.1, "x+1\\over y-1"), fontsize(12pt))
```

image

If not using a Pluto notebook, one needs to add the following javascript in the HTML header
```html
<script type="text/javascript" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_SVG"></script>
```

## Allowing inserting Raw SVG elements
I feel this provides flexibility to make an extension. Although most people do not need this. For example, the following code also renders the Mathjax
```julia
compose(context(), rawsvg(raw"""<foreignObject x="50" y="50" width="100" height="100">
    <div xmlns="http://www.w3.org/1999/xhtml" style="font-family:Times; font-size:15px">
		\(\displaystyle{x+1\over y-1}\)
    </div>
  </foreignObject>"""))
```

# Notes
**Please review this PR carefully because I am not quite familiar with Compose API.**

Reference:
https://stackoverflow.com/questions/15962325/mathjax-inside-svg#:~:text=Currently%2C%20the%20only%20way%20to,'t%20support%20.

### DIFF
--- a/src/Compose.jl
+++ b/src/Compose.jl
@@ -28,7 +28,8 @@ export compose, compose!, Context, UnitBox, AbsoluteBoundingBox,
        vbottom, SVG, SVGJS, PGF, PNG, PS, PDF, draw, pad, pad_inner, pad_outer,
        hstack, vstack, gridstack, LineCapButt, LineCapSquare, LineCapRound,
        CAIROSURFACE, introspect, set_default_graphic_size, set_default_jsmode,
-       boundingbox, Patchable
+       boundingbox, Patchable,
+       rawsvg, mathjax
 
 abstract type Backend end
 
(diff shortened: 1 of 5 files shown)


GiggleLiu:
I find a possible issue, the viewBox is too large for the latex to render correctly sometime. Because MathJax has a minimum superscript/subscript size.

image

GiggleLiu:
I feel this kind of rendering is not perfect. Maybe not a good fit for this package.

PR #426, opened 2021-07-11
Use the html tag for text/html

diegozea:
Hi!

This fixes the interactivity of *Gadfly* plots on *Pluto*, solving the *Gadfly* example in https://github.com/fonsp/Pluto.jl/issues/546 and this [stackoverflow question](https://stackoverflow.com/questions/66297678/using-gadfly-interactively-in-pluto). The fix is similar to the one implemented in https://github.com/kimikage/ProfileSVG.jl/pull/50

**Before** (note that the plot don't react to the mouse)

pluto_no

**After** (grid highlighted because of the interaction, also the question mark appears)

pluto_si

This doesn't break **IJulia**

ijulia_gadfly

nor the example in http://gadflyjl.org/stable/man/backends/#Interactive-SVGs

doc_gadfly

I hope this helps

Best,

### DIFF
--- a/src/svg.jl
+++ b/src/svg.jl
@@ -511,7 +511,12 @@ function show(io::IO, ::MIME"text/html", img::SVG)
     if img.cached_out === nothing
         img.cached_out = String(take!(img.out))
     end
-    write(io, img.cached_out)
+    write(io, 
+        """
+        <html>
+        $(img.cached_out)
+        </html>
+        """)
 end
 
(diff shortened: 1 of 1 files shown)


bjarthur:
thanks!

diegozea:
You're welcome 🙂

diegozea:
Hi! Is it possible to tag a version including this change? Thanks

PR #429, opened 2022-01-05
Revert "Use the html tag for text/html"

fonsp:
This reverts commit 3dbc5a020dbdf300cb0b1c059acbe443676181f1.

The commit was originally motivated by https://github.com/fonsp/Pluto.jl/issues/546 but the issues mentioned here have been solved in Pluto by now.

Using inline HTML instead of a full `<html>` document has much better performance, see https://github.com/JuliaPlots/Plots.jl/pull/3559 for more info

cc @diegozea


fonsp:
Never mind, there are still some issues with script-inside-svg 😬

PR #438, opened 2023-02-20
fix for AbstractString other than String, e.g. InlineString

bjarthur:
need for https://github.com/GiovineItalia/Gadfly.jl/issues/1607

### DIFF
--- a/src/fontfallback.jl
+++ b/src/fontfallback.jl
@@ -96,7 +96,7 @@ function max_text_extents(font_family::AbstractString, size::Measure,
         textheight = 0.0
         for chunk in split(text, '\n')
             chunkheight = glyphheight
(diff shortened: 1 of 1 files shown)


src/fontfallback.jl
```diff
@@ -124,7 +124,7 @@ function text_extents(font_family::AbstractString, size::Measure, texts::Abstrac
         textheight = 0.0
         for chunk in split(text, "\n")
             chunkheight = glyphheight
-            if match(r"<su(p|b)>", text) != nothing
+            if match(r"<su(p|b)>", String(chunk)) != nothing
```
bjarthur, line 127: i'm pretty sure that `text` should be `chunk` here based on the surrounding code.  tests pass either way though.  did not look into this further.

PR #441, opened 2024-03-06
avoid using internal `.data` field

KristofferC:
The previous way is not the correct way of getting a string out of an `IOBuffer` (fails on 1.11 for example)


KristofferC:
cc @bjarthur

bjarthur:
thanks!  can you add the missing right paren?

KristofferC:
classic in-browser editing

PR #444, opened 2025-02-17
Update Compat for Colors.jl to 0.13, bump patch version

zsunberg:
Hello! A few of my packages depend on Compose and Colors. Can we update the compat for Colors to include 0.13? It appears to not break anything.


bjarthur:
thanks!  we should probably bump the minimum required julia version to at least 1.6 if not 1.10 at some point too.  maybe that would fix the macOS CI test failure.

PR #455, opened 2026-05-29
Update to JSON 1

### LINKED ISSUE #449: Tagging a new version & Update JSON
It would be great if a patch release could be triggered that includes #447 . This would unblock some compats for us downstream.

lukem12345:
Close #449 

Updating to JSON.jl 1 caused the test suite to fail with:
```julia
ERROR: LoadError: MethodError: no method matching text_width(::JSON.Object{String, Any}, ::SubString{String}, ::Float64)

Closest candidates are:
  text_width(::Dict, ::AbstractString, ::Float64)
   @ Compose ~/work/Compose.jl/Compose.jl/src/fontfallback.jl:74
```

Looking at the [JSON.jl migration guide](https://juliaio.github.io/JSON.jl/stable/migrate/#Reading-JSON), usage of `Dict` was replaced with `JSON.Object`:
> JSON.Object{String, Any} is now the default type used when parsing instead of Dict{String, Any}; JSON.Object is a drop-in replacement for Dict, supporting the AbstractDict interface, mutation, dot-access (getproperty) to keys, memory and performance benefits for small objects vs. Dict, and preserves the JSON order of keys. For large objects (hundreds or thousands of keys), or to otherwise restore the pre-1.0 behavior, you can do JSON.parse(json; dicttype=Dict{String, Any}).

So, I replaced the `Dict` type specification with `AbstractDict`, since that is the shared supertype of `Dict` and `JSON.Object`, and this solved the immediate issue. A quick check of the `length` of `keys` of `glyphheight` and `glyphwidths` shows `1` and `95` respectively, so the change in performance should be locally OK.

(The testsuite tests with just JSON 1, so automated checks going forward that this package works with < 1 versions of JSON.jl would require some CI/ CD configurations or changes to the test suite.)


bjarthur:
thanks!

